Issue #8, opened 2021-06-17
Get github actions CI working 


brendanheywood:
Example: https://github.com/catalyst/moodle-mod_applaunch/blob/main/.github/workflows/main.yml

brendanheywood:
example:

https://github.com/catalyst/moodle-mod_applaunch/blob/main/.github/workflows/main.yml

nigelchapman-catalyst:
CI Passing. 

![image](https://user-images.githubusercontent.com/46659321/123015262-47d80400-d40b-11eb-84a4-a069ab52119f.png)

Issue #7, opened 2021-06-17
Have an admin option to purge old profiles after X days / weeks


brendanheywood:
* [ ] have an option for a maximum number of profiles (default 100)
* [ ] have an option for a maximum age of profiles ( default 3 months)

nigelchapman-catalyst:
Running php admin/cli/cron.php  is working now... Tested with a 2 hour timeout...

![image](https://user-images.githubusercontent.com/46659321/123200934-1176b380-d4f5-11eb-916d-0493b1a55598.png)

Before:

![image](https://user-images.githubusercontent.com/46659321/123200812-ce1c4500-d4f4-11eb-9cdf-8efcc6633836.png)

After:

![image](https://user-images.githubusercontent.com/46659321/123200892-f6a43f00-d4f4-11eb-9e6d-fdd35b5ec573.png)

brendanheywood:
Moving this back to the backlog as it will need to be redone after the schema change

Issue #2, opened 2021-06-17
Refactor parts of lib.php into auto loaded classes where is makes sense


brendanheywood:

nigelchapman-catalyst:
Reorganisation looks OK; should do a walk-through, lots of little questions

Issue #1, opened 2021-06-11
Add admin setting for sampling rate


brendanheywood:

nigelchapman-catalyst:
![image](https://user-images.githubusercontent.com/46659321/123025874-ace92500-d41e-11eb-91d3-ebb3e39922a2.png)

In and working.

PR #26, opened 2021-06-24
Initial code review

nigelchapman-catalyst:
Code review from 23 Jun; plus tickets #1, #2, #7, #8


index.php
```diff
@@ -157,7 +154,7 @@ function init() {
 
         if (window.excimerData === undefined) {
             setLoading(true);
-            d3.json("json.php?<?php echo $_SERVER['QUERY_STRING'] ?>", function(error, data) {
+            d3.json("flamegraph.json.php?<?php echo $_SERVER['QUERY_STRING'] ?>", function(error, data) {
```
brendanheywood, line 157: using a raw QUERY_STRING feels like a security issue. We do want this to be generic, so maybe the better way is to use new moodle_url() and then use it to mutate the script and then ->out on it

settings.php
```diff
@@ -27,12 +27,13 @@
...
 if ($hassiteconfig) {
 
-    $ADMIN->add('tools', new admin_externalpage(
-        'toolexcimer',
+    $report = new admin_externalpage(
+        'tool_excimer_report',
         get_string('pluginname', 'tool_excimer'),
-        "$CFG->wwwroot/$CFG->admin/tool/excimer/index.php",
+        new moodle_url('/admin/tool/excimer/index.php'),
         'moodle/role:manage'
```
brendanheywood, line 34: this shouldn't use the role:manage. Easiest for now is use ''moodle/site:config' but later we might make our own capability

styles.css
```diff
@@ -0,0 +1,14 @@
+/* tool_excimer styles */
+
+.path-admin-tool-excimer-index #loading {
+  text-align: center;
```
brendanheywood, line 4: I think there is a code standard for 4 spaces for css, think thats the same as php and everything else

version.php
```diff
@@ -23,10 +23,14 @@
...
+$plugin->release = 2021062400;
+
+$plugin->requires = 2019052006;    // Our lowest supported Moodle (3.7.6).
+
+// TODO $plugin->supported = ;     // Available as of Moodle 3.9.0 or later.
+// TODO $plugin->incompatible = ;  // Available as of Moodle 3.9.0 or later.
+
 $plugin->component = 'tool_excimer';
 $plugin->maturity = MATURITY_ALPHA;
 $plugin->release = 'v3.7-r1';
```
brendanheywood, line 36: this is duplicated, remove this line

PR #29, opened 2021-06-28
Profile table

nigelchapman-catalyst:
# TODO.md

[NC 2021-06-28] 

This is a simple MVP that logs all calls and stores profiles when execution
time exceeds a trigger given in admin settings adding similar fields and checks
should be pretty simple. Next steps:

- To find bottlenecks, the flame graphs for the individual profiling should
show time spent in each function, not just the total calls made to a function.

- Need sorting and filtering for profile table on homepage: I've omitted a
basic tablelib implementation here as I haven't gotten it right yet.

- NOTE: admin/cli/cron.php does not cleanup obsolete profiles yet; only calls.

### DIFF
--- a/classes/excimer_call.php
+++ b/classes/excimer_call.php
@@ -16,77 +16,65 @@
 
 namespace tool_excimer;
 
+use ExcimerLog;
 use ExcimerLogEntry;
 use tool_excimer\excimer_helper;
 
 defined('MOODLE_INTERNAL') || die();
 
 /**
- * Manage excimer_log table
+ * Manage excimer_call table
  *
  * @package   tool_excimer
  * @author    Nigel Chapman <nigelchapman@catalyst-au.net>
  * @copyright 2021, Catalyst IT
  * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
-class excimer_log {
+class excimer_call {
 
     /**
      * Save Excimer log to DB.
      *
-     * moodle=# \d mdl_tool_excimer;
-                                       Table "public.mdl_tool_excimer"
-        Column    |         Type          |                           Modifiers
-    --------------+-----------------------+---------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


classes/excimer_call.php
```diff
@@ -123,9 +111,29 @@ public static function summarize() {
...
+     * Select log data by profileid
+     *
+     * @param int $profileid
+     * @return iterable Result set for query
+     */
+    public static function get_profile_data($profileid) {
+        global $DB;
+        $safeprofileid = (int)$profileid;
+        $sql = "
+              SELECT  graphpath,
```
brendanheywood, line 127: tiny code standard: just 1 space between SELECT the second half

classes/excimer_profile.php
```diff
@@ -0,0 +1,157 @@
...
+                debugging('Insert failed for tool_excimer_profile; table missing?');
+            }
+            return $id;
+        } else {
+            return null;
+        }
+    }
+
+    public static function is_cli() {
+        return php_sapi_name() === 'cli';
```
brendanheywood, line 107: there is a moodle define for this

define('CLI_SCRIPT', true);

lib.php
```diff
@@ -97,21 +96,20 @@ function tool_excimer_spool(ExcimerLog $log) {
...
-        $prof->stop();
-        $prof->flush();
-        if (is_iterable($excimerlogentries)) {
-            foreach ($excimerlogentries as $entry) {
-                excimer_log::save_entry($entry);
+    global $DB, $excimerlogs;
+    $isenabled = (bool)get_config('tool_excimer', 'excimerenable');
+    if ($isenabled) {
+        // Running uninstall_package in CLI will error here unless we check.
+        $tableexists = $DB->get_manager()->table_exists('tool_excimer_call');
```
brendanheywood, line 103: would it be better to just wrap this in an exception catch?

Issue #6, opened 2021-06-17
Have admin options to auto save profiles which took more than X seconds


brendanheywood:
Also have a sibling option which is to only keep the top N slowest profiles, so if the latest profile is not as slow as the top Nth profile then don't save it. If it is then save it and garbage collect the slowest one.

nigelchapman-catalyst:
![image](https://user-images.githubusercontent.com/46659321/123026759-0bfb6980-d420-11eb-9dbc-df109db40c21.png)

Added to admin settings; confirmed working (e.g. set to 10 to record all, set to 60000 to record none). 

Second option will wait until profile table added in subsequent tickets.

jaypha:
Some questions about N slowest.
If a profile is added manually by FLAMEME, is it subject to this test, or will it be always saved?
When the setting is changed, is there a callback or event that can be called to trim the profile table if N goes down?
Thanks

brendanheywood:
No it will always be saved

jaypha:
- [ ] Check number of saved profiles when updating settings
- [ ] unit tests

PR #35, opened 2021-12-01
Issue #6: Automatically saving profiles

jaypha:
Supports automatically profiling through config setting.
Will save a profile if longer than the set time (will always save if uses FLAMEME)
Better support for downloading profile table.
Added descriptions to settings.

### DIFF
--- a/classes/helper.php
+++ b/classes/helper.php
@@ -29,14 +29,14 @@ defined('MOODLE_INTERNAL') || die();
 class helper {
 
     /**
-     * Returns a printable string for a calltype value.
+     * Returns a printable string for a script type value.
      *
-     * @param int $calltype
+     * @param int $type
      * @return string
      * @throws \coding_exception
      */
-    public static function scripttypeasstring(int $calltype): string {
-        switch ($calltype) {
+    public static function script_type_display(int $type): string {
+        switch ($type) {
             case profile::SCRIPTTYPE_WEB:
                 return get_string('excimertype_web', 'tool_excimer');
             case profile::SCRIPTTYPE_CLI:
@@ -45,6 +45,26 @@ class helper {
                 return get_string('excimertype_ajax', 'tool_excimer');
             case profile::SCRIPTTYPE_WS:
                 return get_string('excimertype_ws', 'tool_excimer');
+            default:
+                return (string) $type;
+        }
+    }
+
+    /**
+     * Returns a printable string for the logging method.
+     *
+     * @param int $method
+     * @return string
+     * @throws \coding_exception
+     */
+    public static function log_method_display(int $method): string {
+        switch ($method) {
+            case manager::LOGMETHOD_MANUAL:
+                return get_string('excimerlogmethod_manual', 'tool_excimer');
+            case manager::LOGMETHOD_AUTO:
+                return get_string('excimerlogmethod_auto', 'tool_excimer');
+            default:
+                return (string) $method;
         }
     }
 }
--- a/classes/helper.php
+++ b/classes/helper.php
@@ -51,20 +51,20 @@ class helper {
     }
 
     /**
-     * Returns a printable string for the logging method.
+     * Returns a printable string for the profiling reason.
      *
-     * @param int $method
+     * @param int $reason
      * @return string
      * @throws \coding_exception
      */
-    public static function log_method_display(int $method): string {
-        switch ($method) {
-            case manager::LOGMETHOD_MANUAL:
-                return get_string('excimerlogmethod_manual', 'tool_excimer');
-            case manager::LOGMETHOD_AUTO:
-                return get_string('excimerlogmethod_auto', 'tool_excimer');
+    public static function reason_display(int $reason): string {
+        switch ($reason) {
+            case manager::REASON_MANUAL:
+                return get_string('excimerreason_manual', 'tool_excimer');
+            case manager::REASON_AUTO:
+                return get_string('excimerreason_auto', 'tool_excimer');
             default:
-                return (string) $method;
+                return (string) $reason;
         }
     }
 }
--- a/classes/manager.php
+++ b/classes/manager.php
@@ -30,8 +30,9 @@ class manager {
 
     const MANUAL_PARAM_NAME = 'FLAMEME';
 
-    const LOGMETHOD_MANUAL = 0;
-    const LOGMETHOD_AUTO = 1;
+    // Reason for profiling.
+    const REASON_MANUAL = 0;
+    const REASON_AUTO = 1;
 
     const EXCIMER_LOG_LIMIT = 10000;
     const EXCIMER_PERIOD = 0.01;  // Default in seconds; used if config is out of sensible range.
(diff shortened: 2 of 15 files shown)


jaypha:
The checks fail because the unit tests require the Excimer PHP extension to be installed.

jaypha:
The checks will fail if MariaDB 10.6 is used. MariaDB 10.6 treats compressed row format as read-only.

classes/manager.php
```diff
@@ -86,16 +95,58 @@ public static function init(): void {
...
+                    }
+                }
+            }
+        }
+        if ($dowesave) {
+            profile::save($log, $logmethod, (int) $started, $duration);
+        }
+    }
+
+    public static function onnumslowsettingchange(string $name) {
```
brendanheywood, line 145: while vars can't have _ a function name can to be more readable

classes/profile.php
```diff
@@ -97,16 +97,78 @@ public static function getashar($id): string {
...
+
+    /**
+     * @return object
+     * @throws \dml_exception
+     */
+    public static function getfastestautoprofile(): object {
+        global $DB;
+        $sql = "
+            SELECT id, duration
+            FROM {tool_excimer_profiles}
```
brendanheywood, line 119: please indent the left capital works so their right letters are aligned, eg

https://docs.moodle.org/dev/SQL_coding_style#Indentation

classes/profile.php
```diff
@@ -97,16 +97,78 @@ public static function getashar($id): string {
...
+    }
+
+    /**
+     * @param $numtopurge
+     * @throws \dml_exception
+     */
+    public static function purgefastestautoprofiles($numtopurge): void {
+        global $DB;
+        $sql = "
+            DELETE FROM {tool_excimer_profiles}
```
brendanheywood, line 150: please convert this to use one of the dml methods, probably $DB->delete_records_subquery

brendanheywood: also would it be easier to swap this around so the param is the number you want to keep?

Then purgefastestautoprofile could be redundant as you'd just say purge_fastest_autoprofiles($keep = 0)

classes/helper.php
```diff
@@ -46,6 +46,21 @@ public static function scripttypeasstring(int $calltype): string {
             case profile::SCRIPTTYPE_WS:
                 return get_string('excimertype_ws', 'tool_excimer');
         }
+        throw new \coding_exception('Unsupported script type ' . $type);
```
brendanheywood, line 49: if it is just display logic which isn't critical, I'd make this kind of code more forgiving and just print an 'unknown' instead of exploding

classes/helper.php
```diff
@@ -46,6 +46,21 @@ public static function scripttypeasstring(int $calltype): string {
...
+     * @return string
+     */
+    public static function logmethoddisplaytring(int $method): string {
+        switch ($method) {
+            case manager::LOGMETHOD_MANUAL:
+                return get_string('excimerlogmethod_manual', 'tool_excimer');
+            case manager::LOGMETHOD_AUTO:
+                return get_string('excimerlogmethod_auto', 'tool_excimer');
+        }
+        throw new \coding_exception('Unsupported logging method ' . $method);
```
brendanheywood, line 63: same again: if it is just display logic which isn't critical, I'd make this kind of code more forgiving and just print an 'unknown' instead of exploding

settings.php
```diff
@@ -41,24 +43,59 @@
     );
     $ADMIN->add('tools', $settings);
 
-    $settings->add(new admin_setting_configcheckbox(
-        'tool_excimer/excimerenable',
-        new lang_string('excimerenable', 'tool_excimer'),
-        '', 0));
+    if ($ADMIN->fulltree) {
+        $settings->add(
+            new admin_setting_configselect(
```
brendanheywood, line 48: I'm unable to test this, I don't see any of the plugin settings or report links anymore, not sure why

I think I'd prefer to not combine the settings and stick with a single enable but I'd like to test it first and see

settings.php
```diff
@@ -41,24 +43,59 @@
...
+                    manager::LOGMETHOD_AUTO => get_string('excimerlogmethod_auto', 'tool_excimer')
+                ]
+            )
+        );
 
-    $settings->add(new admin_setting_configduration(
-        'tool_excimer/excimerexpiry_s',
-        new lang_string('excimerexpiry_s', 'tool_excimer'),
-        '', WEEKSECS));
+        // TODO make this appear only when logmethod is auto.
```
brendanheywood, line 60: Conditional admin settings are on the spectrum of hard so I'd not attempt this. It is also nice to be able to change settings even when they are not in use

classes/profile.php
jaypha: It is more efficient this way. I can just list them in order and slice from the top. If you prefer to use num-to-keep, I can change it, but it will require extra logic.

classes/manager.php
jaypha: This has been largely implemented.

classes/profile.php
jaypha: This has been done.

classes/profile.php
jaypha: The function has been converted to use DML method.

classes/helper.php
jaypha: Done

classes/helper.php
jaypha: Done

settings.php
jaypha: It has been converted back to a checkbox.

settings.php
jaypha: The TODO comment has been removed.

jaypha:
All the comments raised have been addressed.

index.php
```diff
@@ -65,6 +66,7 @@
 
 $headers = [
     get_string('excimerfield_request', 'tool_excimer'),
+    'Log method',
```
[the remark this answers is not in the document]
jaypha: Oops!

classes/profile.php
```diff
@@ -133,30 +171,30 @@ private static function getparameters(int $type): string {
...
         $flamedatad3 = json_encode(converter::process($flamedata));
-        $type = self::getscripttype();
-        $parameters = self::getparameters($type);
-
+        $type = self::get_script_type();
+        $parameters = self::get_parameters($type);
         return $DB->insert_record('tool_excimer_profiles', [
             'sessionid' => session_id(),
-            'type' => $type,
+            'logmethod' => $logmethod,
```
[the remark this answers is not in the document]
jaypha: profiletype?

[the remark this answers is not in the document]
jaypha: Renamed to 'reason'.

classes/profile_table.php
```diff
@@ -46,7 +50,11 @@ public function col_type(object $record): string {
      * @return string
      */
     public function col_request(object $record): string {
-        return "<a href='/admin/tool/excimer/profile.php?id=$record->id'>$record->request</a>";
+        if ($this->is_downloading()) {
+            return $record->request;
+        } else {
+            return "<a href='/admin/tool/excimer/profile.php?id=$record->id'>$record->request</a>";
```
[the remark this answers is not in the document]
jaypha: done

index.php
```diff
@@ -41,12 +43,10 @@
 
 // TODO support downloading.
```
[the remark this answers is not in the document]
jaypha: done

index.php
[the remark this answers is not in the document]
jaypha: done

settings.php
```diff
@@ -41,24 +43,59 @@
...
+        $setting =
+        new admin_setting_configtext(
+            'tool_excimer/excimernum_slowest',
```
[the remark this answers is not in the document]
jaypha: done

PR #40, opened 2021-12-06
Issue #7: Purging old profiles

jaypha:
Added profile::purge_profiles_before_epoch_time()
Added profile::get_num_profiles()
Added unit test for tool_excimer_profile_test::test_purge_old_profiles()

### DIFF
--- a/classes/profile.php
+++ b/classes/profile.php
@@ -211,7 +211,6 @@ class profile {
     /**
      * Delete profiles created earlier than a given time.
      *
-     * @authpr Nigel Chapman <nigelchapman@catalyst-au.net>
      * @param int $cutoff Epoch seconds
      * @return void
      */
--- a/classes/profile.php
+++ b/classes/profile.php
@@ -96,6 +96,11 @@ class profile {
         return '';
     }
 
+    public static function get_num_profiles(): int {
+        global $DB;
+        return $DB->count_records('tool_excimer_profiles', []);
+    }
+
     /**
      * @return int
      * @throws \dml_exception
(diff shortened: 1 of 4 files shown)


db/tasks.php
```diff
@@ -33,7 +33,7 @@
         'minute' => '0',
         'hour' => '*',
         'day' => '*',
-        'dayofweek' => '*',
-        'month' => '*'
+        'month' => '*',
```
brendanheywood, line 36: This can't hurt, but also this pattern will be wrong in so many places its not really worth worrying about given that it's cosmetic

classes/profile.php
```diff
@@ -201,4 +206,22 @@ public static function save(\ExcimerLog $log, int $reason, int $created, float $
             'flamedatad3' => $flamedatad3
         ]);
     }
+
+
+    /**
+     * Delete profiles created earlier than a given time.
+     *
+     * @authpr Nigel Chapman <nigelchapman@catalyst-au.net>
```
brendanheywood, line 214: Should be you?

brendanheywood: What about the exist delete_before_epoch_time function, should this be removed too or kept?

jaypha: Nigel did write this one. I merely pulled it accross.
I'm hanging on to the excimer_* classes for now. But they will eventually be removed entirely.

db/tasks.php
jaypha: May as well. As you said, it can't hurt.

PR #41, opened 2021-12-06
Issue #7: Stored profile limit

jaypha:
Added 'tool_excimer/excimermaxstored' setting, for an absolute max store.
Added check to is_profiling() to check that limit has not been reached.


classes/manager.php
```diff
@@ -58,8 +58,12 @@ public static function is_flag_set(string $flag): bool {
      * @throws \dml_exception
      */
     public static function is_profiling(): bool {
-        return self::is_flag_set(self::MANUAL_PARAM_NAME) ||
-                (get_config('tool_excimer', 'excimeranableauto'));
+        return (
+            self::is_flag_set(self::MANUAL_PARAM_NAME) ||
+            (bool)get_config('tool_excimer', 'excimeranableauto')
+        ) && (
+            profile::get_num_profiles() < (int)get_config('tool_excimer', 'excimermaxstored')
```
brendanheywood, line 65: I'm not sure if we need this setting at all, given we already have excimernum_slowest. It may make more sense to convert it into a limit just for the manually triggered profiles but I'll mull on this for a bit.

But if we do keep it then I would prefer to not do a db read here. I think it would be more correct to profiling it and then decide after it is done whether it is interesting enough to keep, and then only if it is do the db read to see if it should be kept and push a different older / faster one out of the list.

I was going to leave all the optimization work towards the end but it's worth having my end goal in mind. I don't want this plugin to have any performance impact on a running system, the absolute bare minimum of db or cache reads and ideally no unnecessary writes during a web request that we can't defer to cron or avoid somehow in the first place.

Issue #20, opened 2021-06-21
Add a watch user mode


brendanheywood:
This would be similar to the PROFILEALL mode but is trigger session by session and turned on when a userid / username / email hits one of a specific set of values in a textarea. Possibly also trigger it when any user hits a specific page and from then on

* [  ] store the username for the current user
* [  ] store the session id - partially redacted

jaypha:
So I understand, you currently want userid, username, email and url for triggers. Any others (for now)?
Do you always want the profiling to last for the remainder of the session?
Do you want to be able to combine triggers? e.g. certain user hits a certain url?
Thanks

brendanheywood:
This issue is only a single trigger which is FLAMEALL. It will have the same usage as tool_profiling and FLAMEME, ie it will flip a bit in your session and continuously profile everything in this and future requests until you do FLAMEALLSTOP as a query param

We may add other triggers later such as triggering depending on who the user is, or what script etc, but thats out of scope for this issue.

The existing code is already recording the redacted session id, can you please also record the userid of the logged in user as well. That actually independent of this new trigger and should apply across the board

jaypha:
I think I get it. Someone uses FLAMEALL as a param, then it profiles until FLAMEALLSTOP as a PARAM. Ok?

Actually, the whole session ID is recorded, but redacted on display. I can change it do store it redacted if you want.

brendanheywood:
Correct

Lets split that into https://github.com/catalyst/moodle-tool_excimer/issues/43

PR #44, opened 2021-12-07
Issue #20: FLAMEALL support

jaypha:
Support for parameters FLAMEALL and FLAMEALLSTOP to switch profiling on and off respectively.
Added userid to profile DB table. #20

### DIFF
--- a/classes/manager.php
+++ b/classes/manager.php
@@ -60,12 +60,13 @@ class manager {
      * @return bool True if we have flame all set.
      */
     protected static function is_flame_all(): bool {
+        global $SESSION;
         if (self::is_flag_set(self::FLAME_OFF_PARAM_NAME)) {
-            unset($_SESSION[self::FLAME_ON_PARAM_NAME]);
+            unset($SESSION->toolexcimerflameall);
         } else if (self::is_flag_set(self::FLAME_ON_PARAM_NAME)) {
-            $_SESSION[self::FLAME_ON_PARAM_NAME] = 1;
+            $SESSION->toolexcimerflameall = true;
         }
-        return isset($_SESSION[self::FLAME_ON_PARAM_NAME]);
+        return isset($SESSION->toolexcimerflameall);
     }
 
     /**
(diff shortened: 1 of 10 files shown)


classes/manager.php
```diff
@@ -51,14 +54,29 @@ public static function is_flag_set(string $flag): bool {
...
     }
 
+    /**
+     * Checks flame on/off flags and sets the session value.
+     *
+     * @return bool True if we have flame all set.
+     */
+    protected static function is_flame_all(): bool {
+        if (self::is_flag_set(self::FLAME_OFF_PARAM_NAME)) {
+            unset($_SESSION[self::FLAME_ON_PARAM_NAME]);
```
brendanheywood, line 64: please use the global $SESSION->foo = 'bar'; instead of $_SESSION which is the moodle code standard way of doing things

db/install.xml
```diff
@@ -15,6 +15,7 @@
         <FIELD NAME="request" TYPE="char" LENGTH="256" NOTNULL="true" SEQUENCE="false" COMMENT="URL, command, or similar identifying string"/>
         <FIELD NAME="parameters" TYPE="char" LENGTH="256" NOTNULL="true" SEQUENCE="false" COMMENT="Request variables or command arguments"/>
         <FIELD NAME="sessionid" TYPE="char" LENGTH="256" NOTNULL="true" SEQUENCE="false" COMMENT="Session ID"/>
+        <FIELD NAME="userid" TYPE="int" LENGTH="11" NOTNULL="true" SEQUENCE="false" COMMENT="User ID"/>
```
brendanheywood, line 18: can you please add a fk relation ship for this eg

```
        <KEY NAME="userid" TYPE="foreign" FIELDS="userid" REFTABLE="user" REFFIELDS="id"/>

```

jaypha: Added.

jaypha:
Added display for user in both table and profile.

classes/profile_table.php
```diff
@@ -97,4 +127,18 @@ public function col_created(object $record): string {
...
+     * Displays the full name of the user.
+     *
+     * @param object $record
+     * @return string
+     */
+    public function col_user(object $record): string {
+        if ($record->userid == 0) {
+            return '-';
+        } else {
+            return fullname($record);
```
brendanheywood, line 141: can you please link this to the user profle

auth/catadmin/classes/form/admin_report.php:131:            $fullname = \html_writer::link('/user/profile.php?id=' . $user->id, $fullname);

templates/flamegraph.mustache
```diff
@@ -29,13 +29,15 @@
...
     * sessionid - The session ID
+    * userid - ID of the user who loaded the page
     * type - The script type
     * created - Unix timestamp of profile creation
     * duration - Profile duration in seconds.
     * parameters - Parameter string
     * resposnecode - HTTP response code
     * cookies - Is cookies enabled?
     * buffering - Is buffering enabled?
+    * fullname - full nane of user represented by userid.
```
brendanheywood, line 40: typo

classes/profile_table.php
jaypha: done in both the table and the template.

Issue #42, opened 2021-12-06
Remove redundant 'excimer' prefixes from admin settings


brendanheywood:
This overlaps a lot with https://github.com/catalyst/moodle-tool_excimer/issues/38

eg

https://github.com/catalyst/moodle-tool_excimer/blob/main/settings.php#L55-L57

```
                'tool_excimer/excimersample_ms',
                get_string('excimerperiod_ms', 'tool_excimer'),
                get_string('excimerperiod_ms_desc', 'tool_excimer'),             
```

could be simpler as just:

```
                'tool_excimer/sample_ms',
                get_string('period_ms', 'tool_excimer'),
                get_string('period_ms_desc', 'tool_excimer'),             
```

Issue #38, opened 2021-12-03
Remove redundant excimer prefix from lang strings


brendanheywood:
Might be best to do this combined with

https://github.com/catalyst/moodle-tool_excimer/issues/42

PR #48, opened 2021-12-08
Issue #38: Remove redundant excimer prefixes.

jaypha:
Removed 'excimer' prefix from language strings and config names.
Includes issue #42
Fix bug where cookies and buffering data was not saved properly.


classes/profile.php
```diff
@@ -201,8 +201,8 @@ public static function save(\ExcimerLog $log, int $reason, int $created, float $
             'duration' => $duration,
             'request' => $_SERVER['PHP_SELF'] ?? 'UNKNOWN',
             'parameters' => $parameters,
-            'cookies' => !defined(NO_MOODLE_COOKIES),
-            'buffering' => !defined(NO_OUTPUT_BUFFERING),
+            'cookies' => !defined('NO_MOODLE_COOKIES') || !NO_MOODLE_COOKIES,
```
brendanheywood, line 204: isn't this boolean reversed?

ie if cookies = 1 should that mean we did have a cookie?

Issue #55, opened 2021-12-09
Fix crumb trail when drilling down into a profile so you can get back to the index page


brendanheywood:
similar to the core profiles

![image](https://user-images.githubusercontent.com/187449/145331859-fbd7f677-a68e-4785-b3dc-a7c5483e8844.png)

Issue #39, opened 2021-12-06
Break up the reports into different reports for slowest, most recent


brendanheywood:
* [x] add each report to the admin tree
* [x] move the reports to the debugging top level menu, then under a second level heading called 'excimer' (this is to workaround totara nav which messes with reports)
* [x] report for most recent 
* [x] report for slowest
* [x] add a shared sub-tab widget to both reports so you can swap between them easier

PR #57, opened 2021-12-12
Issue #39: New report pages

jaypha:
Added pages for recent and slowest profile reports.
Moved them to the Development Category.
Created profile_table_page class to handle common code.

There is a display issue with the breadcrumbs, but since that is to be fixed in issue #55, I thought to leave it for now.

### DIFF
--- a/classes/profile_table_page.php
+++ b/classes/profile_table_page.php
@@ -38,8 +38,6 @@ class profile_table_page {
 
         $context = \context_system::instance();
 
-        admin_externalpage_setup('tool_excimer_report_' . $report);
-
         $PAGE->set_context($context);
         $PAGE->set_url($url);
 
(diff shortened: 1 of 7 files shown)


classes/profile_table_page.php
```diff
@@ -0,0 +1,99 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_excimer;
+
+class profile_table_page {
```
brendanheywood, line 19: This refactor is sufficiently close to the output api that I think its worth doing it all the way:

https://docs.moodle.org/dev/Output_API

specifically:

1) move it to the \output\ name space
2) optionally turn it into a render (happy to defer this, not critical at all)

PR #69, opened 2021-12-14
Detect and prevent profiling the class does not exist #68

### LINKED ISSUE #68: Enabling enable_auto setting causes exception to appear on every page load
Enable auto profiling tool_excimer | enable_auto 

Enabling the above setting on a clean install causes the following warning to appear on subsequent page loads.

- Installed excimer using PECL steps.
- The excimer plugin was installed from an upgrade process of an existing site.

```
Exception calling 'tool_excimer_after_config'

    line 44 of /admin/tool/excimer/lib.php: call to tool_excimer\manager::init()
    line 1073 of /lib/setup.php: call to tool_excimer_after_config()
    line 82 of /config.php: call to require_once()
    line 26 of /user/files.php: call to require()
```

keevan:
image

Closes #68

### DIFF
--- a/lang/en/tool_excimer.php
+++ b/lang/en/tool_excimer.php
@@ -30,6 +30,7 @@ $string['report_slowest'] = 'Profile list - slowest';
 $string['report_recent'] = 'Profile list - recent';
 
 // Settings.
+$string['here'] = 'here';
 $string['general_settings'] = 'General settings';
 $string['general_settings_desc'] = 'Settings related to all profiles';
 $string['auto_settings'] = 'Auto profiling settings';
@@ -48,7 +49,7 @@ $string['request_ms'] = 'Minimum request duration (milliseconds)';
 $string['request_ms_desc'] = 'Record a profile only if it runs at least this long.';
 $string['num_slowest_by_page'] = 'Max to save by page';
 $string['num_slowest_by_page_desc'] = 'Only the N slowest profiles will be kept for each script page.';
-$string['noexcimerprofiler'] = 'ExcimerProfiler class does not exist so profiling cannot continue.';
+$string['noexcimerprofiler'] = 'ExcimerProfiler class does not exist so profiling cannot continue. Please check the installation instructions {$a}.';
 
 // Tabs.
 $string['recent'] = 'Recent';
--- a/lang/en/tool_excimer.php
+++ b/lang/en/tool_excimer.php
@@ -48,6 +48,7 @@ $string['request_ms'] = 'Minimum request duration (milliseconds)';
 $string['request_ms_desc'] = 'Record a profile only if it runs at least this long.';
 $string['num_slowest_by_page'] = 'Max to save by page';
 $string['num_slowest_by_page_desc'] = 'Only the N slowest profiles will be kept for each script page.';
+$string['noexcimerprofiler'] = 'ExcimerProfiler class does not exist so profiling cannot continue.';
 
 // Tabs.
 $string['recent'] = 'Recent';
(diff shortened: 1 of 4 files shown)


settings.php
```diff
@@ -58,11 +58,14 @@
...
-            'tool_excimer/general',
-            get_string('general_settings', 'tool_excimer'),
-            get_string('general_settings_desc', 'tool_excimer'),
-        ));
+        $warntext = '';
+        if (!class_exists('ExcimerProfiler')) {
+            $warntext  .= $OUTPUT->notification(get_string('noexcimerprofiler', 'tool_excimer'));
+        }
+        $warntext .= get_string('general_settings_desc', 'tool_excimer');
+        $config = \tool_objectfs\local\manager::get_objectfs_config();
```
brendanheywood, line 66: cut and paste error?

lang/en/tool_excimer.php
```diff
@@ -48,6 +48,7 @@
 $string['request_ms_desc'] = 'Record a profile only if it runs at least this long.';
 $string['num_slowest_by_page'] = 'Max to save by page';
 $string['num_slowest_by_page_desc'] = 'Only the N slowest profiles will be kept for each script page.';
+$string['noexcimerprofiler'] = 'ExcimerProfiler class does not exist so profiling cannot continue.';
```
brendanheywood, line 51: Can you add some words to help someone fix the issue, even if it's just a link to:

https://github.com/catalyst/moodle-tool_excimer#installation

lib.php
```diff
@@ -36,11 +36,9 @@
  * eliminates the need for an auto_prepend_file/auto_append_file.
  */
 function tool_excimer_after_config() {
-
     // TODO Temp ref: https://docs.moodle.org/dev/Login_callbacks#after_config
     // TODO Do we want to check if in upgrade/install etc.
```
brendanheywood, line 40: Can you please test and protect against a few more edge case states, eg during an install or upgrade. I'm fairly sure the _after_config callback doesn't fire at all under these so we get it free, in which just refine the todo

lang/en/tool_excimer.php
keevan: Sure thing. I had it that way because though running through the install instructions I only resolved it after a container restart. Hopefully not always the case, but I'll add it to maybe a troubleshooting section as well.

lib.php
keevan: Okay, will check these states

jaypha: It does call it when uninstalling. When I run the uninstall CLI program, it complains that it cannot save the profile (because the tables don't exist). So it is likely true for upgrading as well, but I never checked. It is not an issue for installing as auto profiling is switched off by default.

settings.php
keevan: Very much so, thanks for spotting that :)

lib.php
keevan: I tested an exception throw here during upgrade, which appears it does get called during upgrade. Having the extra check here seems to prevent the issue from showing during an upgrade.
```
++ Exception calling 'tool_excimer_after_config' ++
* line 1073 of /lib/setup.php: call to tool_excimer_after_config()
* line 67 of /config.php: call to require_once()
* line 41 of /admin/cli/upgrade.php: call to require()
-->tool_excimer
++ 2021112901: Success (0.02 seconds) ++
++ Success (0.12 seconds) ++
== Setting new default values ==
Command line upgrade from 3.9.10+ (Build: 20211022) (2020061510.06) to 3.9.10+ (Build: 20211022) (2020061510.06) completed successfully.
```

keevan: That said, did we want profiling to happen as part of these processes?

brendanheywood: from a business point of view I'd love to be able to see a profile of upgrades so and optimize them. So ideally yes. But also happy to just make it not explode for the moment and tackle this later.

PR #70, opened 2021-12-14
Align request path prefixes

### LINKED ISSUE #51: align cli vs web path prefix, and confirm paths when installed under directories
Profiles scripts don't have a /, but web ones do:

image

* [ ] Remove the / from the web ones.
* [ ] Also check what happens if moodle is installed under /somedir/admin/cli etc. I think in all cases remove the wwwroot prefix 
* [ ] Remove wwwroot from paths inside the profile flame

keevan:
image

* [x] Remove the / from the web ones.
* [x] Remove wwwroot from paths inside the profile flame

> Also check what happens if moodle is installed under /somedir/admin/cli etc. I think in all cases remove the wwwroot prefix 

I believe based on the above adjustments, it shouldn't really care about where it's installed. If it's installed under a weird path that replicates the directory structure of the moodle site, that's an edge case which I don't think would happen in a normal sane environment, unless moodle starts changing their pathing to end up with something like `/path/to/site/var/log` and the user installs it under `/var/log`

Closes #51

### DIFF
--- a/classes/profile.php
+++ b/classes/profile.php
@@ -159,11 +159,15 @@ class profile {
      * @throws \dml_exception
      */
     public static function save(\ExcimerLog $log, int $reason, int $created, float $duration): int {
-        global $DB, $USER;
+        global $DB, $USER, $SCRIPT;
         $flamedata = trim(str_replace("\n;", "\n", $log->formatCollapsed()));
         $flamedatad3 = json_encode(converter::process($flamedata));
         $type = self::get_script_type();
         $parameters = self::get_parameters($type);
+
+        // If set, it will trim off the leading '/' to normalise web & cli requests.
+        $request = isset($SCRIPT) ? ltrim($SCRIPT, '/') : 'UNKNOWN';
+
         return $DB->insert_record('tool_excimer_profiles', [
             'sessionid' => substr(session_id(), 0, 10),
             'reason' => $reason,
@@ -172,7 +176,7 @@ class profile {
             'method' => $_SERVER['REQUEST_METHOD'] ?? '',
             'created' => $created,
             'duration' => $duration,
-            'request' => $_SERVER['PHP_SELF'] ?? 'UNKNOWN',
+            'request' => $request,
             'parameters' => $parameters,
             'cookies' => !defined('NO_MOODLE_COOKIES') || !NO_MOODLE_COOKIES,
             'buffering' => !defined('NO_OUTPUT_BUFFERING') || !NO_OUTPUT_BUFFERING,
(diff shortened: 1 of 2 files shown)


classes/profile.php
```diff
@@ -159,11 +159,19 @@ private static function get_parameters(int $type): string {
...
+
+        // Remove full pathing to dirroot and only keep pathing from site root (non-issue in most sane cases).
+        $flamedata = str_replace($CFG->dirroot . DIRECTORY_SEPARATOR, '', $flamedata);
+
         $flamedatad3 = json_encode(converter::process($flamedata));
         $type = self::get_script_type();
         $parameters = self::get_parameters($type);
+
+        // If set, it will trim off the leading '/' to normalise web & cli requests.
+        $request = isset($_SERVER['PHP_SELF']) ? ltrim($_SERVER['PHP_SELF'], '/') : 'UNKNOWN';
```
brendanheywood, line 173: does this work when we are installed inside a folder? wonder if we used $ME it would help

brendanheywood: actually I think we want $SCRIPT which does the logic for us

keevan: Tested both $ME and $SCRIPT

I think $SCRIPT is closer to what we want (example output below)
```
/admin/cli/upgrade.php
/admin/tool/excimer/profile.php
```
We don't have to do any ltrim, as both will have the `/` included in their result. However that means we will keep the `/`, assuming that's okay.

keevan:
.

classes/profile.php
brendanheywood: the minimalist in me wants to ditch the / in all cases

keevan:
```suggestion
        $request = isset($SCRIPT) ? ltrim($SCRIPT, '/') : 'UNKNOWN';
```
¿Por qué no los dos? (Let's do both)

Also do you know if $SCRIPT will ever be UNKNOWN (not set)? It feels like it should always link to an existing script / file and the cases where it's "UNKNOWN" isn't made clear.

Traced it back to here: https://github.com/catalyst/moodle-tool_excimer/commit/9302a864a9491a02f5c739bf491998f486d08d1f#diff-2301d91ad2a2fdf01fbfb4be839244907712b0c1713b457e79e39a8c6d07aab4

brendanheywood: I think they should be rare:

https://github.com/moodle/moodle/blob/master/lib/setuplib.php#L907-L912

jaypha: I'm thinking that if the core code supports the possibility, then we should too.

keevan: That sounds reasonable. I've kept the check and  moved the string to a class constant, and fixed up affected unit tests.

Issue #47, opened 2021-12-08
Have ability to delete profiles


brendanheywood:
We probably want this to be contextual and in a few places
* [x] Have a bin icon next to each profile, use action_link with a js 'confirm action' 
* [x] Have a reset all button
* [ ] when drilling into a single script, have a 'delete everything about this script' button

Issue #60, opened 2021-12-13
'Slowest' report should group by script and sort by max() of each script


brendanheywood:
* [x] group by script

* [x] sort by worst stored
* [x] show a col with how many profiles we have for this script
* [x] created makes no sense, but possibly show the min / max of the date as a range instead, Earliest, Latest as two cols
* [ ] user might not make sense. But try this as a a group aggregation, eg 'Brendan (10), Jason (5)' meaning brendan did 10 profiles and jason did 5. This isn't critical so if its a pita just drop it
* [ ] response code, would be nice to keep it and do as an aggregation similar to above
* [x] make a new 'script detail' page which takes a script without wwwroot prefix as a param
* [x] this shows the subset of all profiles related to this script. This could possibly be one of the existing pages but with an optional ?script=foo filter

jaypha:
Will probably want a way to expand out any script group and list the profiles.

brendanheywood:
Yup added it to the list above

jaypha:
- [ ] Delete by script

jaypha:
I don't think having user and responsecode display this way is possible without a very complex SQL query. It would require returning a variable amount of values, which I do not know how to do.

The alternative would be to use `GROUP BY request, userid, responsecode`, and then reprocess the data in code before giving it to the table.

jaypha:
user id and response code aggregates are to be removed due to version support issues.

PR #74, opened 2021-12-14
Issue #47: Delete Profiles

jaypha:
Issue #47
Added ability to delete profiles.
A 'Delete All' button added to profile list page.
A trash can icon added to profile row to delete individual profiles.

### DIFF
--- a/classes/profile_table.php
+++ b/classes/profile_table.php
@@ -38,11 +38,11 @@ class profile_table extends \table_sql {
         'user',
         'responsecode',
         'referer',
-        'delete'
+        'delete',
     ];
 
     const NOSORT_COLUMNS = [
-        'delete'
+        'delete',
     ];
 
     public function __construct($uniqueid) {
@@ -166,14 +166,10 @@ class profile_table extends \table_sql {
      */
     public function col_delete(object $record) {
         global $OUTPUT;
-        $deleteurl = new \moodle_url('/admin/tool/excimer/delete.php', [ 'deleteid' => $record->id, 'sesskey' => sesskey() ]);
-        $link = new \action_link(
-                $deleteurl,
-                '',
-                new \confirm_action('Confirm?'),
-                null,
-                new \pix_icon('t/delete', get_string('deleteprofile', 'tool_excimer'))
-        );
+        $deleteurl = new \moodle_url('/admin/tool/excimer/delete.php', ['deleteid' => $record->id, 'sesskey' => sesskey()]);
+        $confirmaction = new \confirm_action(get_string('deleteprofilewarning', 'tool_excimer'));
+        $deleteicon = new \pix_icon('t/delete', get_string('deleteprofile', 'tool_excimer'));
+        $link = new \action_link($deleteurl, '', $confirmaction, null,  $deleteicon);
         return $OUTPUT->render($link);
     }
 }
--- a/classes/profile_table.php
+++ b/classes/profile_table.php
@@ -37,7 +37,12 @@ class profile_table extends \table_sql {
         'parameters',
         'user',
         'responsecode',
-        'referer'
+        'referer',
+        'delete'
+    ];
+
+    const NOSORT_COLUMNS = [
+        'delete'
     ];
 
     public function __construct($uniqueid) {
@@ -46,6 +51,10 @@ class profile_table extends \table_sql {
         $headers = [];
         foreach (self::COLUMNS as $column) {
             $headers[] = get_string('field_' . $column, 'tool_excimer');
+        }
+
+        foreach (self::NOSORT_COLUMNS as $column) {
+            $this->no_sorting($column);
         }
 
         $this->set_sql(
@@ -145,5 +154,26 @@ class profile_table extends \table_sql {
                 return \html_writer::link('/user/profile.php?id=' . $record->userid, $fullname);
             }
         }
+    }
+
+    /**
+     * Display for the delete column
+     *
+     * @param object $record
+     * @return mixed
+     * @throws \coding_exception
+     * @throws \moodle_exception
+     */
+    public function col_delete(object $record) {
+        global $OUTPUT;
+        $deleteurl = new \moodle_url('/admin/tool/excimer/delete.php', [ 'deleteid' => $record->id, 'sesskey' => sesskey() ]);
+        $link = new \action_link(
+                $deleteurl,
+                '',
+                new \confirm_action('Confirm?'),
+                null,
+                new \pix_icon('t/delete', get_string('deleteprofile', 'tool_excimer'))
+        );
+        return $OUTPUT->render($link);
     }
 }
--- a/classes/profile_table.php
+++ b/classes/profile_table.php
@@ -38,11 +38,11 @@ class profile_table extends \table_sql {
         'user',
         'responsecode',
         'referer',
-        'delete',
+        'actions',
     ];
 
     const NOSORT_COLUMNS = [
-        'delete',
+        'actions',
     ];
 
     public function __construct($uniqueid) {
@@ -157,14 +157,14 @@ class profile_table extends \table_sql {
     }
 
     /**
-     * Display for the delete column
+     * Display for the action icons
      *
      * @param object $record
      * @return mixed
      * @throws \coding_exception
      * @throws \moodle_exception
      */
-    public function col_delete(object $record) {
+    public function col_actions(object $record) {
         global $OUTPUT;
         $deleteurl = new \moodle_url('/admin/tool/excimer/delete.php', ['deleteid' => $record->id, 'sesskey' => sesskey()]);
         $confirmaction = new \confirm_action(get_string('deleteprofilewarning', 'tool_excimer'));
--- a/delete.php
+++ b/delete.php
@@ -1,0 +1,55 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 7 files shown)


jaypha:
The delete by script feature can be handled as a part of issue #60.

classes/profile_table.php
```diff
@@ -146,4 +155,25 @@ public function col_user(object $record): string {
...
+     * Display for the delete column
+     *
+     * @param object $record
+     * @return mixed
+     * @throws \coding_exception
+     * @throws \moodle_exception
+     */
+    public function col_delete(object $record) {
+        global $OUTPUT;
+        $deleteurl = new \moodle_url('/admin/tool/excimer/delete.php', [ 'deleteid' => $record->id, 'sesskey' => sesskey() ]);
```
keevan, line 169:
```suggestion
        $deleteurl = new \moodle_url('/admin/tool/excimer/delete.php', ['deleteid' => $record->id, 'sesskey' => sesskey()]);
```
Please kill the surrounding whitespace

classes/profile_table.php
```diff
@@ -146,4 +155,25 @@ public function col_user(object $record): string {
...
+        $link = new \action_link(
+                $deleteurl,
+                '',
+                new \confirm_action('Confirm?'),
```
keevan, line 173: This should be a lang string (get_string call). I believe one exists for 'Confirm' under `lang/en/moodle.php`

jaypha:
Discovered that 'delete all' is still displayed even when the list is empty.

classes/profile_table.php
```diff
@@ -37,7 +37,12 @@ class profile_table extends \table_sql {
         'parameters',
         'user',
         'responsecode',
-        'referer'
+        'referer',
+        'delete'
```
keevan, line 41: Please add a trailing comma to the last item in the list, this makes it easier to extend and a cleaner diff.

>  (with special attention to the last line having a trailing comma to extend the list of items later with a cleaner diff) 

See: https://docs.moodle.org/dev/Coding_style#Arrays

classes/profile_table.php
```diff
@@ -37,7 +37,12 @@ class profile_table extends \table_sql {
...
+    ];
+
+    const NOSORT_COLUMNS = [
+        'delete'
```
keevan, line 45: Same as above please

classes/profile_table.php
```diff
@@ -146,4 +155,25 @@ public function col_user(object $record): string {
...
+    public function col_delete(object $record) {
+        global $OUTPUT;
+        $deleteurl = new \moodle_url('/admin/tool/excimer/delete.php', [ 'deleteid' => $record->id, 'sesskey' => sesskey() ]);
+        $link = new \action_link(
+                $deleteurl,
+                '',
+                new \confirm_action('Confirm?'),
+                null,
+                new \pix_icon('t/delete', get_string('deleteprofile', 'tool_excimer'))
+        );
```
keevan, line 176: Perhaps instead of having this split into multiple lines such as this, it might be clearer to have something similar to this:
```
        $confirmaction = new \confirm_action(/* fix with lang string */);
        $deleteicon = new \pix_icon('t/delete', get_string('deleteprofile', 'tool_excimer'));
        $link = new \action_link($deleteurl, '', $confirmaction, null,  $deleteicon);
```

delete.php
```diff
@@ -0,0 +1,55 @@
...
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+use tool_excimer\profile;
```
keevan, line 17: Symbol 'profile' is declared but not used. Please remove it. Also if there was an include, it would typically fall below the phpdoc below with the file description and author/copyright etc.

delete.php
```diff
@@ -0,0 +1,55 @@
...
+require_once(__DIR__ . '/../../../config.php');
+
+require_once($CFG->libdir.'/adminlib.php');
+
+require_admin();
+
+$returnurl = optional_param('returnurl', '', PARAM_LOCALURL);
+if ($returnurl == '') {
+    $returnurl = get_local_referer(false);
+}
```
keevan, line 37:
```suggestion
$returnurl = optional_param('returnurl', get_local_referer(false), PARAM_LOCALURL);
```
Would this work instead?

delete.php
```diff
@@ -0,0 +1,55 @@
...
+
+    $deleteid = optional_param('deleteid', 0, PARAM_INT);
+
+    if ($deleteid) {
+        $DB->delete_records('tool_excimer_profiles', [ 'id' => $deleteid ]);
+        redirect($returnurl, get_string('profiedeleted', 'tool_excimer'));
+    }
+}
+
+redirect($returnurl);
```
keevan, line 55: I think related things should be grouped together, but that might just be me. Also, I believe it's best practice to declare all your request variables up front at the top, so anyone jumping into the file knows what to expect and what they can use, and it avoids another potential declaration if it was used in another nested block.

With that said, assuming `$deleteall` and `$deleteid` are defined at the top, I suggest the following instead.

```suggestion
// Add a comment here
if (!confirm_sesskey()) {
    redirect($returnurl);
}

// Add a comment here
if ($deleteall) {
    $DB->delete_records('tool_excimer_profiles');
    redirect($returnurl, get_string('allprofiesdeleted', 'tool_excimer'));
}

// Add a comment here
if ($deleteid) {
    $DB->delete_records('tool_excimer_profiles', ['id' => $deleteid]);
    redirect($returnurl, get_string('profiedeleted', 'tool_excimer'));
}
```
Also note that I removed the whitespace around the array contents in what was L50 (previously). I believe having things too deeply nested makes it hard to grok, so this should be easier to understand.

lang/en/tool_excimer.php
```diff
@@ -68,6 +68,7 @@
 $string['field_referer'] = 'Referer';
 $string['field_cookies'] = 'Cookies Enabled';
 $string['field_buffering'] = 'Buffering Enabled';
+$string['field_delete'] = '';
```
keevan, line 71: Added but not used? Please remove or clarify where this is needed. Also it's empty, I don't think this should be here at all in this case.

profile.php
```diff
@@ -66,6 +70,10 @@
 $data['reason_display'] = function($text, $render) {
     return helper::reason_display((int)$render($text));
 };
+$data['delete_button_display'] = function($text, $render) use ($deletebutton) {
+    global $OUTPUT;
+    return $OUTPUT->render($deletebutton);
+};
```
keevan, line 76: If this doesn't need text or render, should this be a direct declaration instead? (Mustache template may also need to be updated)

```suggestion
$data['delete_button_display'] = $OUTPUT->render($deletebutton);
```

lang/en/tool_excimer.php
```diff
@@ -83,3 +84,10 @@
 $string['reason_auto'] = 'Auto';
 $string['reason_flameall'] = 'Flame All';
 
+// Miscellaneous.
+$string['deleteall'] = 'Delete All';
+$string['deleteallwarning'] = 'This will remove ALL stored profiles. Continue?';
+$string['deleteprofile'] = 'Delete profile';
+$string['deleteprofilewarning'] = 'This will remove the profile. Continue?';
+$string['allprofiesdeleted'] = 'All profiles have been deleted';
+$string['profiedeleted'] = 'Profile Deleted';
```
keevan, line 93: Typically all sentences should end with a fullstop. Please add these.

(This does not need to be added for action buttons)

keevan (commented):
I've left some feedback on the current state of the patch. I have not yet functionally tested this PR yet.

delete.php
jaypha: Yes. Yes it would.

lang/en/tool_excimer.php
jaypha: It's added to the table as part of a loop. If it is not there, then some means of determining whether or not there is a heading to display needs to be implemented. I figured this is the simplest.

classes/profile_table.php
```diff
@@ -146,4 +155,21 @@ public function col_user(object $record): string {
...
+
+    /**
+     * Display for the delete column
+     *
+     * @param object $record
+     * @return mixed
+     * @throws \coding_exception
+     * @throws \moodle_exception
+     */
+    public function col_delete(object $record) {
```
brendanheywood, line 167: Not at action for now, just more of a heads up there is a defacto pattern in moodle to have an 'action' column and throw a bunch of icons like delete, details, up, down etc all in the one place. We can refactor this later when we add more actions

delete.php
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @package   tool_excimer
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2021, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+require_once(__DIR__ . '/../../../config.php');
+
+require_once($CFG->libdir.'/adminlib.php');
```
brendanheywood, line 28: do you need this here?

lang/en/tool_excimer.php
brendanheywood: Actually seeing this makes me think this should be renamed to 'Actions' see previous comment

An example in core to copy:

With a menu:
http://moodle.local/admin/tool/analytics/index.php

With icons:
http://moodle.local/admin/searchareas.php

PR #80, opened 2021-12-15
Add pathinfo to profiles

### LINKED ISSUE #76: The query path is not being stored in the profile
Just noticed we missing some crucial parts of the url for urls which use PATH_INFO after the .php bit:

/pluginfile.php/1/core_admin/logocompact/300x300/1639358903/moodle-icon.jpg

### LINKED ISSUE #82: Opening profile on page 2+ causes section error
This is most likely caused by incorrectly handling the return url based on
local referer + basename splitting, as params weren't considered.

Repro steps:
1 - Navigate to page 2 of the list of profiles
2 - Open any of the profiles from this page
3 - See 'section error' message

image

keevan:
This should record the full details for a pluginfile.php request so it can be reconstructed if needed.

Changes highlighted below:

Profile changes:
- add pathinfo and parameters to request text displayed (no truncate, full url displayed)
- parameters will still exist but will show as not-encoded / clear for readers
- referer will be a clickable link

image

List changes: 
- add pathinfo and parameters to request text displayed, 
- removed parameters column as a result of the above
- truncated display to 100 characters
- title attribute will show full request if curious

image

Adjusted to 100 characters - word-break: break-all

image

Closes #76
Closes #82

### DIFF
--- a/classes/profile_table.php
+++ b/classes/profile_table.php
@@ -56,7 +56,7 @@ class profile_table extends \table_sql {
         }
 
         $this->set_sql(
-            '{tool_excimer_profiles}.id as id, reason, scripttype, request, created, duration, parameters, responsecode,
+            '{tool_excimer_profiles}.id as id, reason, scripttype, method, request, pathinfo, created, duration, parameters, responsecode,
                      referer, userid, lang, firstname, lastname, firstnamephonetic, lastnamephonetic, middlename, alternatename',
             '{tool_excimer_profiles} LEFT JOIN {user} on ({tool_excimer_profiles}.userid = {user}.id)',
             '1=1'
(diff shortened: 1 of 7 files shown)


keevan:
Revised changes:

image

image

Want me to add a `Full Request` field or just push them into one field (Request)? (Thinking of the where the grouping could happen later on)

classes/profile_table.php
```diff
@@ -97,11 +97,11 @@ public function col_scripttype(object $record): string {
      */
     public function col_request(object $record): string {
         if ($this->is_downloading()) {
-            return $record->request;
+            return $record->request . $record->pathinfo;
```
brendanheywood, line 100: I think the request columns should be:

1) visibly just the full url, as close as possible to what the user originally saw
2) except with the wwwroot removed
3) the linked url will be the full url
4) it should also include the params, the params column can be removed
5) the visible url may be truncated down to say 100 chars (look at html_to_text to help with this)

brendanheywood: It's probably worth having a helper somewhere to rebuild the url, we will use this in a few places

lang/en/tool_excimer.php
```diff
@@ -63,6 +63,7 @@
 $string['field_user'] = 'User';
 $string['field_duration'] = 'Duration';
 $string['field_request'] = 'Request';
+$string['field_pathinfo'] = 'Path Info';
```
brendanheywood, line 66: don't think we need this

templates/flamegraph.mustache
```diff
@@ -65,6 +66,10 @@
             <th>{{#str}} field_request, tool_excimer {{/str}}</th>
             <td>{{request}}</td>
         </tr>
+        <tr>
+            <th>{{#str}} field_pathinfo, tool_excimer {{/str}}</th>
```
brendanheywood, line 70: same as the table, lets just have a single 'request url' which all the same reused logic as in the first table

classes/profile_table.php
jaypha: You can also look at shorten_text (in lib/moodlelib.php), which is also available as a built in mustache helper.

keevan: I've updated the main PR description with the changes - visuals included.

Issue #59, opened 2021-12-13
Cosmetic random bits


brendanheywood:
* [x] remove params from top level table
* [x] remove referer from top level table
* [x] if duration is > 1 minute round to nearest second
* [x] if duration is < 10 second round to 3 decimal places
* [x] date format '9/12/21, 19:22'  => '9 Dec 21, 19:22'
* [x] add table-sm to make it more compact
* [x] show the response codes wrapped in a bootstrap labels, 2xx in success, 3xx in info, 4xx and 5xx in warning / error

jaypha:
I'll also add these
[] Fix breadcrumb displaying incorrect link when going to a profile from the 'recent' list.
[] Have delete code use require_sesskey()
[] Move profile settings to the Development/Excimer block.

PR #81, opened 2021-12-16
Issue #59: Cosmetic fixes

jaypha:
Issue #59
Some breadcrumb fixes.
Moved settings link to development/excimer
Changed delete.php to use require_sesskey().
Removed 'parameter' and 'referer' columns from profile tables.
Added 'table-sm' class to profile tables.
Response codes are displayed as badges.
Duration and created values have new formats.

### DIFF
--- a/classes/helper.php
+++ b/classes/helper.php
@@ -29,6 +29,16 @@ defined('MOODLE_INTERNAL') || die();
 class helper {
 
     /**
+     * Maps HTTP status codes to css badges.
+     */
+    const STATUS_BADGES = [
+        2 => 'badge-success',
+        3 => 'badge-info',
+        4 => 'badge-warning',
+        5 => 'badge-error',
+    ];
+
+    /**
      * Returns a printable string for a script type value.
      *
      * @param int $type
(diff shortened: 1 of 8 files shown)


classes/helper.php
```diff
@@ -69,5 +79,28 @@ public static function reason_display(int $reason): string {
...
+    }
+
+    /**
+     * Returns HTTP status as a badge.
+     *
+     * @param int $status
+     * @return string
+     */
+    public static function http_status_display(int $status): string {
+        $spanclass = 'badge ' . self::STATUS_BADGES[$status / 100];
```
brendanheywood, line 103: I think this is breaking on a CLI script which has a null status

I think in the case of cli scripts we can do the same stuff, ie 0 = success and non zero is error. I'm not sure if we are actually storing cli return values or if that's actually possible until the script exists so just make this work

classes/helper.php
```diff
@@ -69,5 +79,28 @@ public static function reason_display(int $reason): string {
...
-}
 
+    /**
+     * Returns a formatted time duration in m:s.ms format.
+     * @param float $duration
+     * @return string
+     * @throws \Exception
+     */
+    public static function duration_display(float $duration): string {
+        $ms = round($duration * 1000, 0);
```
brendanheywood, line 90: I think there are some edge cases with this not working, I think use sprintf would be easy

image

Also can you please right align this column

classes/helper.php
brendanheywood: Can you also most this column to the far left before the request

PR #83, opened 2021-12-16
Add support for multiple reasons per profile entry

### LINKED ISSUE #64: Allow any given profile to be stored for multiple reasons
eg a profile might have been triggered anyway because it was slow, but it was also triggered using FLAMEME so record them both.

Probably need a bit mask or something to make this simple in the db. We will want this to be fairly snappy to query on if we decide to ramp up to collecting large numbers of profiles

keevan:
- Simplify output (using implode).
- Simplify storage, removing unnecessary flag setting and checking
- Did NOT touch existing purge tasks, as anything not purely captured
  via automatic means should not be automatically purged even if it met
  the conditions.
- phpcs: Fix trailing whitespace.
- Fix array whitespacing for more consistency.
- phpcs: Fix long line > 132 characters for set_sql call.

Preview:

image

Closes #64

### DIFF
--- a/classes/manager.php
+++ b/classes/manager.php
@@ -32,10 +32,24 @@ class manager {
     const FLAME_ON_PARAM_NAME = 'FLAMEALL';
     const FLAME_OFF_PARAM_NAME = 'FLAMEALLSTOP';
 
-    // Reason for profiling.
-    const REASON_MANUAL = 0;
-    const REASON_AUTO = 1;
-    const REASON_FLAMEALL = 2;
+    /** Reason - MANUAL - Profiles are manually stored for the request using FLAMEME as a page param. */
+    const REASON_MANUAL   = 0b0001;
+
+    /** Reason - AUTO - Set when conditions are met and these profiles are automatically stored. */
+    const REASON_AUTO     = 0b0010;
+
+    /** Reason - FLAMEALL - Toggles profiling for all subsequent pages, until FLAMEALLSTOP param is passed as a page param. */
+    const REASON_FLAMEALL = 0b0100;
+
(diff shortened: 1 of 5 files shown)


classes/manager.php
```diff
@@ -126,18 +126,18 @@ public static function on_flush(\ExcimerLog $log, float $started): void {
         $stopped = microtime(true);
         $duration = $stopped - $started;
 
+        $reason = 0;
```
brendanheywood, line 129: Maybe have reason none = 0 as const

classes/profile.php
```diff
@@ -262,10 +262,10 @@ public static function purge_fastest_by_page(int $numtokeep): void {
      */
     public static function purge_fastest(int $numtokeep): void {
         global $DB;
-        $numtopurge = $DB->count_records('tool_excimer_profiles', ['reason' => manager::REASON_AUTO ]) - $numtokeep;
+        $numtopurge = $DB->count_records('tool_excimer_profiles', ['reason' => manager::REASON_AUTO]) - $numtokeep;
```
brendanheywood, line 265: Should there be some DB->sql_bit_or or similar in here?

classes/manager.php
keevan: Added now, but since this should NEVER be a reason to store the profile, I've skipped adding a langstring for it, as it should never be stored and thus displayed.

keevan:
Changes from review:
- Add unit test to check for reasons being removed, which also checks on last reason removal for record deletion/non-existence
- Updated purge tasks to only remove the REASON_AUTO (will be renamed to REASON_SLOW as part of another PR)

PR #92, opened 2021-12-17
Add params to the request field similar to GET requests #87

### LINKED ISSUE #86: Bug with query params missing from http POST
query params can be present on all http types, post, put, delete etc. The display logic should not use the http method only the presence of query params

### LINKED ISSUE #87: Cli profiles don't have command line arguments, should be treated in a similar way to path info or query params

keevan:
Few screenshots of changed display for CLI commands

Profile

image

List 

image

CLI list

image

CLI profile

image

**Note: The example won't work as a copy and paste due to the fact it's using `\` and it won't be properly escaped e.g. in bash without wrapping the value in quotes or using double back slash.**

Closes #87 
Closes #86

### DIFF
--- a/classes/profile_table.php
+++ b/classes/profile_table.php
@@ -112,6 +112,7 @@ class profile_table extends \table_sql {
             if ($record->scripttype == profile::SCRIPTTYPE_CLI) {
                 // For CLI scripts, request should look like `command.php --flag=value` as an example.
                 $separator = ' ';
+                $record->parameters = escapeshellcmd($record->parameters);
             } else {
                 // For GET requests, request should look like `myrequest.php?myparam=1` as an example.
                 $separator = '?';
--- a/classes/profile_table.php
+++ b/classes/profile_table.php
@@ -108,15 +108,25 @@ class profile_table extends \table_sql {
      */
     public function col_request(object $record): string {
         $displayedrequest = $record->request . $record->pathinfo;
-        if ($record->method === 'GET' && !empty($record->parameters)) {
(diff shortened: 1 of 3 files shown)


brendanheywood:
Can we escape the output with cmdescapeargs and friends ?

classes/profile_table.php
```diff
@@ -62,7 +61,6 @@ public function __construct($uniqueid) {
             '1=1'
         );
         $this->define_columns(self::COLUMNS);
-        $this->column_class('responsecode', 'text-right');
```
brendanheywood, line 65: Why not keep this column on its own? It's visually hard up against it so it's not much different real estate wise, except it is right aligned

keevan: image

Re-added the column to the list page

classes/profile_table.php
```diff
@@ -108,15 +108,25 @@ public function col_scripttype(object $record): string {
      */
     public function col_request(object $record): string {
         $displayedrequest = $record->request . $record->pathinfo;
-        if ($record->method === 'GET' && !empty($record->parameters)) {
-            $displayedrequest .= '?' . urldecode($record->parameters);
+        if (!empty($record->parameters)) {
+            if ($record->scripttype == profile::SCRIPTTYPE_CLI) {
+                // For CLI scripts, request should look like `command.php --flag=value` as an example.
+                $separator = ' ';
```
brendanheywood, line 114: we can escape this with escapeshellcmd so you can cut and paste it into the cli and have it still work.

keevan: image

Output

PR #107, opened 2021-12-21
Issue #15: Save partial profiles for long executing scripts.

### LINKED ISSUE #15: If we have really long requests, then periodically persist the partial profile every N seconds, default to 10
Mark the profile as in progress

Still only do a partial save if we have already passed the normal thresholds for interest.

Lets say there is a cron task which is kill -9'ed or a really long running web request which never completes and the container gets reaped. In these cases the php shutdown handler may not run so we may not get a proper profile saved. To protect against this we should have options to incrementally save a partial profile. The logic will be very similar to #6 but we don't need to save very often, just the mere fact that a profile didn't finish is enough without needing the full details of it. So it could persist once when it hits some threshold (eg configured at 30 seconds) and then re-persist every 100 time slices or throttle back as it gets longer and longer

If we are inside a transaction need special handling, simplest is just don't save yet. But we could open a second db connection which we then close again on the spot

This is somewhat at odds with the 'do no harm' principle, so maybe we can do a single save when we first hit the limit, and then slowly throttle back the saves. Need to tune in real life.

jaypha:
resolves #15
Setup ExcimerTimer to call a function for every N seconds.
Added a setting to set N.
Saving a paritial stores the record ID, so that subsequent saves
will override it.


jaypha:
TODO - unit tests

jaypha:
I have been able to get an alternate DB connection to work for transactions in regular execution. However, unit tests fail massively as a consequence. A call to preventResetByRollback() is needed to stop the unit tests from using transactions.

classes/profile.php
```diff
@@ -186,38 +194,68 @@ public static function save(\ExcimerLog $log, int $reason, int $created, float $
...
-            'contenttypekey' => $contenttypekey,
-            'contenttypecategory' => $contenttypecategory,
-        ]);
+
+        $intrans = $DB->is_transaction_started();
+
+        if ($intrans) {
+            $cfg = $DB->export_dbconfig();
+            $db2 = \moodle_database::get_driver_instance($cfg->dbtype, $cfg->dblibrary);
+            $db2->connect($cfg->dbhost, $cfg->dbuser, $cfg->dbpass, $cfg->dbname, $cfg->prefix, $cfg->dboptions);
```
brendanheywood, line 203: can you add some protection is the super worst case where for some reason we can't connect. In this case just don't do the partial save and gracefully let the script continue.

This might happen when the db is getting overloaded and we max out connections and we don't want to escalate things. Please add a comment to this effect too

PR #115, opened 2021-12-24
Issues 72,89,108: Add new data fields

### LINKED ISSUE #72: Detect and store the versions of code against a profile
Not clear exactly which version, it might be the core version as well as the component version. But a profile might touch a range of different plugins so this could get messy.

At minimum store something such as $CFG->allversionshash which may not be an opaque value but at least we can see when code has changed. Ideally something tied to the code itself not just the version (hard).

### LINKED ISSUE #89: Store user agent for request

### LINKED ISSUE #108: Store the pid and hostname for a profile

jaypha:
resolves #72
resolves #89
resolves #108

Added:
- Process ID
- User agent
- Host name
- Version hash
These are all stored, and displayed in the profiel detail page.


keevan (approved):
Looks good to me.

That said, not sure if this resolves #72 completely. I believe Brendan wanted some more than just the allversionhash so that issue should probably stay open once this is merged.

PR #119, opened 2021-12-28
Issue #99: Delete by filter

### LINKED ISSUE #99: Delete all on 'slowest' tab, with a script selected, should delete just this script
Might be clearer to just add a second button which is exactly the same as the bin icon on the 'slowest grouped' tab

jaypha:
closes #99
Added ability to delete according to a filter.
Added extra delete button to display table for deleting by filter.

### DIFF
--- a/classes/profile_table_page.php
+++ b/classes/profile_table_page.php
@@ -62,7 +62,7 @@ class profile_table_page {
                 echo $output->render($deletebutton);
             }
             $filters = $table->get_filters();
-            if (count($filters)) {
+            if (!empty($filters)) {
                 $deleteurl = new \moodle_url('/admin/tool/excimer/delete.php', ['filter' => json_encode($filters)]);
                 $deletebutton = new \single_button($deleteurl, get_string('deleteprofiles_filter', 'tool_excimer'));
                 $deletebutton->add_confirm_action(get_string('deleteprofiles_filter_warning', 'tool_excimer'));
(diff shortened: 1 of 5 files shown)


delete.php
```diff
@@ -46,10 +47,13 @@
     redirect($returnurl, get_string('profiledeleted', 'tool_excimer'));
 }
 
-// Delete profiles for a script.
-if ($deletescript) {
-    $DB->delete_records('tool_excimer_profiles', ['request' => $deletescript]);
-    redirect($returnurl, get_string('profilesdeleted', 'tool_excimer'));
+// Delte profiles according to a filter value.
```
keevan, line 50: Small typo
```suggestion
// Delete profiles according to a filter value.
```

classes/profile_table_page.php
```diff
@@ -61,6 +61,13 @@ public static function display(profile_table $table, \moodle_url $url): void {
                 $deletebutton->add_confirm_action(get_string('deleteallwarning', 'tool_excimer'));
                 echo $output->render($deletebutton);
             }
+            $filters = $table->get_filters();
+            if (count($filters)) {
```
keevan, line 65:
```suggestion
            if (!empty($filters)) {
```
Since the count itself is not important, we just want to know if there are filters applied at all or not.

keevan (commented):
Aside from these small comments, I tested the functionality and it seems to be working okay.

PR #122, opened 2021-12-29
Optimise profiler storage based on cached durations (#121, #106)

### LINKED ISSUE #106: Profile up to the cache limit and store no more than the least impactful case, based on the REASON
e.g. for REASON_SLOW (or currently named as REASON_AUTO):

if we are profiling requests which are long than 100 seconds, and keeping 50 profiles, and then lets say our quota is full, lets look at the fastest stored profile (the least impactful case) and it might have a time of 107 seconds. We can cache this 107 limit and now safely ignore all requests faster than 107 seconds instead of 100 seconds. If we change the settings, or if we delete any profiles at all, then reset this cache to be safe.

### LINKED ISSUE #121: Profile up to the cached limit, storing profiles slower than the fastest profile for the request/page and reason
Likewise (to https://github.com/catalyst/moodle-tool_excimer/issues/106) if we are only storing 5 of each type of profile then we can cache the slowly ratcheting up lower limit on a per script basis. So if a given request is fairly fast then it only needs to check the cache above, and only then if it passes will we check this second cache.

keevan:
- Caches the min duration per page and per reason (or profile type)
- Uses this cache to adjust the min duration required for it to be
  stored automatically when profiling, currently REASON_SLOW (or AUTO)
  is supported, but built it so other reasons can be supported as well.
- Add cache definition to tool_excimer
- Cache busting callback added on profile save, on certain setting
  updates, and when removing reasons via the purge task.
- Optimised some of the checks added, adding examples to the function
  docs.

Closes #121 Closes #106

### DIFF
--- a/classes/manager.php
+++ b/classes/manager.php
@@ -217,7 +217,7 @@ class manager {
             $result[$cachefield] = $minduration;
             $cache->set($cachekey, $result);
         }
-        return $result[$cachefield];
+        return (float)$result[$cachefield];
     }
 
     /**
@@ -252,7 +252,7 @@ class manager {
             $result = (end($resultset)->min_duration ?? 0) * 1000;
             set_config($cachekey, $result, 'tool_excimer');
         }
-        return $result;
+        return (float)$result;
     }
 
     /**
@@ -299,6 +299,20 @@ class manager {
         return true;
     }
 
+    /**
+     * Clears the plugin cache for keys used for the provided reasons
+     *
+     * @param int $reason bitmap of reason(s)
+     */
+    public static function clear_min_duration_cache_for_reason(int $reason): void {
+        foreach (self::REASONS as $basereason) {
+            if ($reason & $basereason) {
+                // Clear the plugin config cache for this profile's reason.
+                $cachekey = 'profile_type_' . $basereason . '_min_duration_ms';
+                unset_config($cachekey, 'tool_excimer');
+            }
+        }
+    }
 
     /**
      * Called when the Excimer log flushes.
--- a/classes/manager.php
+++ b/classes/manager.php
@@ -273,14 +273,14 @@ class manager {
         // First, check against the trigger_ms value to ensure it meets the
         // minimum required duration for the profile to be considered slow.
         $triggerms = get_config('tool_excimer', 'trigger_ms');
-        if ($triggerms && $duration < $triggerms) {
+        if ($triggerms && $duration <= $triggerms) {
             return false;
         }
 
         // If a min duration exists, it means the quota is filled, and only
         // profiles slower than the fastest stored profile should be stored.
         $minduration = self::get_min_duration_for_reason(self::REASON_SLOW);
-        if ($minduration && $duration < $minduration) {
+        if ($minduration && $duration <= $minduration) {
             return false;
         }
 
@@ -290,7 +290,7 @@ class manager {
         // profiles slower than the fastest stored profile should be stored.
         $request = profile::get_request();
         $requestminduration = self::get_min_duration_for_request_and_reason($request, self::REASON_SLOW);
-        if ($requestminduration && $duration < $requestminduration) {
+        if ($requestminduration && $duration <= $requestminduration) {
             return false;
         }
 
--- a/classes/manager.php
+++ b/classes/manager.php
@@ -168,7 +168,8 @@ class manager {
         if (isset($SESSION->toolexcimerflameall)) {
             $reason |= self::REASON_FLAMEALL;
         }
-        if (($duration * 1000) >= self::get_min_duration_for_reason_slow()) {
+
+        if (self::is_considered_slow($duration * 1000)) {
             $reason |= self::REASON_SLOW;
         }
         return $reason;
@@ -176,6 +177,9 @@ class manager {
 
     /**
      * Returns the minimum duration for profiles matching this reason and page/request.
+     *
+     * Cost: 1 cache read (ideally)
+     * Otherwise: 1 cache read, 1 DB read and 1 cache write.
      *
      * @param  int $reason - the profile type or REASON_*
      * @return float duration (in milliseconds) of the fastest profile for a given reason and request/page.
@@ -188,10 +192,11 @@ class manager {
 
         // Grab the fastest profile for this page/request, and use that as
         // the lower boundary for any new profiles of this page/request.
-        $cachekey = "profile_type_$reason" . "_page_$request" . '_min_duration_ms';
-        $cache = \cache::make('tool_excimer', 'timings');
+        $cachekey = $request;
+        $cachefield = "min_duration_for_reason_$reason";
+        $cache = \cache::make('tool_excimer', 'request_metadata');
         $result = $cache->get($cachekey);
-        if ($result === false) {
+        if ($result === false || !isset($result[$cachefield])) {
             // NOTE: Opting to query this way instead of using MIN due to
             // the fact valid profiles will be added and the limits will be
             // breached for 'some time'. This will keep the constraints as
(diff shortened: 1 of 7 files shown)


keevan:
Note: This adds up to 2 extra db calls for every request to check the quotas, namely `has_filled_page_and_reason_quota` and failing that `has_filled_reason_quota`.

I was wondering if this should be cached, e.g. if the quota is reached, it would probably stay capped until purged by a cleanup script, and even then it would probably still stay capped.

classes/profile.php
```diff
@@ -269,6 +281,11 @@ public static function save(\ExcimerLog $log, int $reason, int $created, float $
         if ($intrans) {
             $db2->dispose();
         }
+
+        // Clear the profile timings cache on insert/update of a profile.
+        $cache = \cache::make('tool_excimer', 'timings');
```
brendanheywood, line 286: isn't this too broad and we only want to delete a single cache item not the whole lot?

classes/profile.php
```diff
@@ -315,6 +334,13 @@ public static function remove_reason(array $profiles, int $reason): void {
         if (!empty($idstodelete)) {
             list($insql, $inparams) = $DB->get_in_or_equal($idstodelete);
             $DB->delete_records_select('tool_excimer_profiles', 'id ' . $insql, $inparams);
+            $updateordelete = true;
+        }
+
+        if ($updateordelete) {
+            // Clear the profile timings cache on insert/update of a profile.
+            $cache = \cache::make('tool_excimer', 'timings');
+            $cache->purge();
```
brendanheywood, line 343: ditto, too broad?

classes/manager.php
```diff
@@ -163,12 +168,190 @@ public static function get_reasons(float $duration): int {
...
         return $reason;
     }
 
+    /**
+     * Returns the minimum duration for profiles matching this reason and page/request.
+     *
+     * @param  int $reason - the profile type or REASON_*
+     * @return float duration (in milliseconds) of the fastest profile for a given reason and request/page.
+     */
+    public static function get_min_duration_for_request_and_reason(string $request, int $reason): float {
```
brendanheywood, line 184: At the moment there is only one reason (slow) for a given url. But lets say there ends up being 10, wouldn't this be better if there was a single cache item for a given request and all the metadata for a url was in that cache item rather than doing 10 cache reads? Some of the things we might stash in here might not be timings so I'd call it something a bit more generic eg: make('tool_excimer', 'request_metadata');

classes/manager.php
```diff
@@ -163,12 +168,190 @@ public static function get_reasons(float $duration): int {
...
+    public static function get_min_duration_for_reason(int $reason): float {
+        global $DB;
+
+        $reasonstr = self::REASON_STR_MAP[$reason];
+        $quota = (int) get_config('tool_excimer', "num_$reasonstr");
+        $cache = \cache::make('tool_excimer', 'timings');
+        // Grab the fastest profile across the slow profiles, and use that
+        // as the lower boundary for any new profiles.
+        $cachekey = "profile_type_$reason" . '_min_duration_ms';
+        $result = $cache->get($cachekey);
```
brendanheywood, line 233: This should only be a single number across the system (per reason), so I would not use the cache api this, I'd store it directly using get / set_config and then it is simply there already without another cache read as get_config loads all config for a plugin in one call.

classes/manager.php
```diff
@@ -163,12 +168,190 @@ public static function get_reasons(float $duration): int {
...
+            $sql = "SELECT duration as min_duration
+                      FROM {tool_excimer_profiles}
+                     WHERE $reasons != ?
+                           AND request = ?
+                  ORDER BY duration DESC
+                     ";
+            $resultset = $DB->get_records_sql($sql, [
+                self::REASON_NONE,
+                $request,
+            ], 0, $pagequota);
```
brendanheywood, line 210: I think this can be simpler and we only grab the Nth record, and if there is < N records then we won't get a record and we can cache 0 instead. By not ever storing anything other than the exact Nth record we can make the logic simpler in get_min_duration_for_reason_slow

classes/manager.php
```diff
@@ -163,12 +168,190 @@ public static function get_reasons(float $duration): int {
...
+     */
+    public static function get_min_duration_for_reason_slow(): float {
+        // Quota for this page, for this profile type (e.g. REASON_SLOW) has been reached.
+        $request = profile::get_request();
+
+        // Get the cached timings for the fastest of the stored profiles, to
+        // ensure anything faster than this does not get stored iif the quota is
+        // reached. This cache should be reset when, a new profile is
+        // stored/deleted, or settings have changed.
+        if (self::has_filled_page_and_reason_quota($request, self::REASON_AUTO)) {
```
brendanheywood, line 339: Would this be simpler and use lets lookups if you did get_min_duration_for_reason() first (which should not require any extra read, see above) and only then in the small number of cases that exceed that then check the per requests cached limit.

I also don't think we need to check the quota in either case. If we only ever store the exact Nth quota we implicitly get this for free

classes/manager.php
keevan: Yes that makes sense.

classes/manager.php
keevan: Very nice catch here. Yes we can ignore the quota lookups with this as well.

classes/manager.php
keevan: Yes based on the above feedback, this would reduce the overall cost of lookups assuming the 
- request quota is not reached
- and the profile is rejected because it is not slow enough to be counted as a profile to be stored.

Because now the cost of the system (per reason) quota is free/zero as mentioned above, if stored via get/set_config which is always loaded fully apparently. Inversely, If the costs of the 2 checks were the same, it would have made more sense to check the conditions required to whether or not it should be stored.

I wonder if we can cheat and also store the 'quota filled' request durations in get/set_config, and as mentioned above do the check if this key exists. The gain here is that we can avoid another single cache read e.g. for a slower than usual request endpoint (slower than the fastest stored profile for the reason, but faster than the fastest stored profile for the reason AND request), where the both quotas have been filled up and the profiler is deciding if the profile is worth adding, at the expense of having durations for N requests (or an arbitrary cap) stored in the plugin config as well.

settings.php
```diff
@@ -38,6 +38,19 @@
     $ADMIN->add('tool_excimer_reports', $settings);
 
     if ($ADMIN->fulltree) {
+
+        // Ensure if particular setting(s) are updated, the cache for profile
+        // timings is cleared, at at most once.
```
jaypha, line 43: Typo here

classes/manager.php
```diff
@@ -163,12 +168,137 @@ public static function get_reasons(float $duration): int {
...
+                      FROM {tool_excimer_profiles}
+                     WHERE $reasons != ?
+                  ORDER BY duration DESC
+                     ";
+            $resultset = $DB->get_records_sql($sql, [
+                self::REASON_NONE,
+            ], $quota - 1, 1); // Will fetch the Nth item based on the quota.
+            // Cache the results in milliseconds (avoids recalculation later).
+            $result = (end($resultset)->min_duration ?? 0) * 1000;
+            set_config($cachekey, $result, 'tool_excimer');
```
jaypha, line 253: You set the config here if it has not already been set. However, I do not see anywhere where this config gets reset. Once set the first time, it doesn't change. Even after the max number gets reached, I can still add more, and debugger runs shows the config value to remain '0'.

jaypha (changes requested):
While the cache gets reset within remove_reason(), it does not get reset inside purge_profiles_before_epoch_time() or delete.php.

classes/manager.php
keevan: Thanks for spotting this, yes I forgot to clear the cache for `set_config` calls in the appropriate places.

jaypha (approved):

PR #126, opened 2021-12-29
Issue #65: compress flame data.

### LINKED ISSUE #65: Consider compressing the data in the profiles using gzip

jaypha:
resolves #65
Added code to compress/decompress data stored in database.
Upgrade code includes compressing existing data.
Note: Uses default compression level.

### DIFF
--- a/db/install.xml
+++ b/db/install.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<XMLDB PATH="admin/tool/excimer/db" VERSION="20211224" COMMENT="XMLDB file for Moodle admin/tool/excimer"
+<XMLDB PATH="admin/tool/excimer/db" VERSION="20211229" COMMENT="XMLDB file for Moodle admin/tool/excimer"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:noNamespaceSchemaLocation="../../../../lib/xmldb/xmldb.xsd"
 >
@@ -28,7 +28,7 @@
         <FIELD NAME="versionhash" TYPE="char" LENGTH="256" NOTNULL="true" SEQUENCE="false" COMMENT="All version hash"/>
         <FIELD NAME="datasize" TYPE="int" LENGTH="11" NOTNULL="true" SEQUENCE="false" COMMENT="Size of the profile data (in bytes)"/>
         <FIELD NAME="numsamples" TYPE="int" LENGTH="11" NOTNULL="true" SEQUENCE="false" COMMENT="Number of samples taken"/>
-        <FIELD NAME="flamedatad3" TYPE="text" NOTNULL="false" SEQUENCE="false" COMMENT="JSON structure compatible with d3-flame-graph."/>
+        <FIELD NAME="flamedatad3" TYPE="binary" NOTNULL="false" SEQUENCE="false" COMMENT="compressed JSON structure compatible with d3-flame-graph."/>
         <FIELD NAME="contenttypecategory" TYPE="char" LENGTH="30" NOTNULL="false" SEQUENCE="false" COMMENT="core_filetypes' resolved string value"/>
         <FIELD NAME="contenttypekey" TYPE="char" LENGTH="30" NOTNULL="false" SEQUENCE="false" COMMENT="core_filetypes's key from the types available."/>
         <FIELD NAME="contenttypevalue" TYPE="char" LENGTH="30" NOTNULL="false" SEQUENCE="false" COMMENT="Raw content type as returned from the content-type response header, if available"/>
--- a/db/install.xml
+++ b/db/install.xml
@@ -28,7 +28,7 @@
         <FIELD NAME="versionhash" TYPE="char" LENGTH="256" NOTNULL="true" SEQUENCE="false" COMMENT="All version hash"/>
         <FIELD NAME="datasize" TYPE="int" LENGTH="11" NOTNULL="true" SEQUENCE="false" COMMENT="Size of the profile data (in bytes)"/>
         <FIELD NAME="numsamples" TYPE="int" LENGTH="11" NOTNULL="true" SEQUENCE="false" COMMENT="Number of samples taken"/>
-        <FIELD NAME="flamedatad3" TYPE="binary" NOTNULL="false" SEQUENCE="false" COMMENT="compressed JSON structure compatible with d3-flame-graph."/>
+        <FIELD NAME="flamedatad3" TYPE="binary" NOTNULL="false" SEQUENCE="false" COMMENT="Compressed JSON structure compatible with d3-flame-graph."/>
         <FIELD NAME="contenttypecategory" TYPE="char" LENGTH="30" NOTNULL="false" SEQUENCE="false" COMMENT="core_filetypes' resolved string value"/>
         <FIELD NAME="contenttypekey" TYPE="char" LENGTH="30" NOTNULL="false" SEQUENCE="false" COMMENT="core_filetypes's key from the types available."/>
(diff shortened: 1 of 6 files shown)


keevan:
I ran the upgrade locally, which failed because my local versions were higher, but this would have happened for any script that hadn't upgrade yet. I wonder if it's easier to start clean again (upgrade.php) since it's still pre-alpha
```
Default exception handler: Cannot downgrade tool_excimer from 2021123000 to 2021122900. Debug:
Error code: cannotdowngrade
* line 720 of /lib/upgradelib.php: downgrade_exception thrown
* line 1918 of /lib/upgradelib.php: call to upgrade_plugins()
* line 193 of /admin/cli/upgrade.php: call to upgrade_noncore()

!!! Cannot downgrade tool_excimer from 2021123000 to 2021122900. !!!
!!
Error code: cannotdowngrade !!
!! Stack trace: * line 720 of /lib/upgradelib.php: downgrade_exception thrown
* line 1918 of /lib/upgradelib.php: call to upgrade_plugins()
* line 193 of /admin/cli/upgrade.php: call to upgrade_noncore()
 !!
Exception ignored in shutdown function Closure::__invoke: Error writing to database (ERROR:  invalid byte sequence for encoding "UTF8": 0x9c
INSERT INTO mdl_tool_excimer_profiles (sessionid,reason,pathinfo,scripttype,userid,method,created,finished,duration,request,parameters,cookies,buffering,responsecode,referer,pid,hostname,useragent,versionhash,datasize,numsamples,flamedatad3,contenttypevalue,contenttypekey,contenttypecategory) VALUES($1,$2,$3,$4,$5,$6,$7,$8,$9,$10,$11,$12,$13,$14,$15,$16,$17,$18,$19,$20,$21,$22,$23,$24,$25) RETURNING id
[array (
  'sessionid' => '',
  'reason' => 2,
  'pathinfo' => '',
  'scripttype' => 1,
  'userid' => 0,
  'method' => '',
  'created' => 1640832221,
  'finished' => 1640832222,
  'duration' => 1.1982588768005371,
  'request' => 'admin/cli/upgrade.php',
  'parameters' => '--non-interactive --allow-unstable',
  'cookies' => 0,
  'buffering' => 1,
  'responsecode' => 0,
  'referer' => '',
  'pid' => 7721,
  'hostname' => 'e40712d597ed',
  'useragent' => '',
  'versionhash' => 'e660de2b319b2e53b164876ba34e776ede073176',
  'datasize' => 958,
  'numsamples' => 12,
(t-l(Y~7D}̄%9>}w|蟟9'xXێ8*ТZl!G-`*^M"*%ue""_G-\'%ǣf+DO   ;*3%}C.nY#Zc I:[G>.d\'CΛv' . "\0" . 'vȷ :/
                                                     I>꼓??Ļ|WA̴x3;YOW쑣$BouvJIryE]z
                                                                                  HlL7IZr
                                                                                         pQ9i4Cz$Y!]bT\\Μ;*
><8;g4		G}Je' . "\0" . 'E77PIߥi@ϙ)?V*Wz,?`rUBg.:[KH΍T2T`Fvi
ݛK1U³fk=^צN8B0;WOp-םXҭ\\l+Se/1S4b,[tǎkdAe\\	-ʫR*/⽧\'q;	\'	\'ngELL~}(\'|>kG}~}΄o*R@0(?#F
}ii-
5eSj: K0J:' . "\0" . ':L6M"j4
                             [{@LI\\' . "\0" . 'O(P!zv}1]&J=@ b-Yg߂t	7y7bRzOyϐ[FQq7<3Ѽ
                                                                                         -/wy*ġ
                                                                                               fioX㓵~Dh$+Ox]n{~/+FϽ[irupLub)_=\\',
  'contenttypevalue' => NULL,
  'contenttypekey' => NULL,
  'contenttypecategory' => NULL,
)])
```

db/install.xml
```diff
@@ -28,7 +28,7 @@
         <FIELD NAME="versionhash" TYPE="char" LENGTH="256" NOTNULL="true" SEQUENCE="false" COMMENT="All version hash"/>
         <FIELD NAME="datasize" TYPE="int" LENGTH="11" NOTNULL="true" SEQUENCE="false" COMMENT="Size of the profile data (in bytes)"/>
         <FIELD NAME="numsamples" TYPE="int" LENGTH="11" NOTNULL="true" SEQUENCE="false" COMMENT="Number of samples taken"/>
-        <FIELD NAME="flamedatad3" TYPE="text" NOTNULL="false" SEQUENCE="false" COMMENT="JSON structure compatible with d3-flame-graph."/>
+        <FIELD NAME="flamedatad3" TYPE="binary" NOTNULL="false" SEQUENCE="false" COMMENT="compressed JSON structure compatible with d3-flame-graph."/>
```
keevan, line 31: Probably a capital to start off the comment.
```suggestion
        <FIELD NAME="flamedatad3" TYPE="binary" NOTNULL="false" SEQUENCE="false" COMMENT="Compressed JSON structure compatible with d3-flame-graph."/>
```

db/upgrade.php
```diff
@@ -159,5 +159,29 @@ function xmldb_tool_excimer_upgrade($oldversion) {
...
+        // Changing type of field flamedatad3 on table tool_excimer_profiles to binary.
+        $table = new xmldb_table('tool_excimer_profiles');
+        $field = new xmldb_field('flamedatad3', XMLDB_TYPE_BINARY, null, null, null, null, null, 'numsamples');
+
+        // A text field cannot be directly converted into a byte field.
+        $records = $DB->get_records('tool_excimer_profiles', null, '', 'id, flamedatad3');
+
+        // Launch change of type for field flamedatad3.
+        $dbman->drop_field($table, $field);
+        $dbman->add_field($table, $field);
```
keevan, line 173: To avoid issues like the one I mentioned earlier in this PR ... `ERROR:  invalid byte sequence for encoding "UTF8"...`

I would have tried adding a new field for this data (e.g. flamedatad3gz), pointed the inserts for new profiles at this new flamedatad3 field, probably converted the existing profiles similar to how you have done here, as you still want to drop the column afterwards.

That said, it probably happened because of the state of my env as opposed to it being something that should be changed here.

db/upgrade.php
```diff
@@ -159,5 +159,29 @@ function xmldb_tool_excimer_upgrade($oldversion) {
...
+
+        // We need to convert any exiting records to store the compressed data.
```
keevan, line 175: `exiting` => `existing`

db/upgrade.php
```diff
@@ -159,5 +159,29 @@ function xmldb_tool_excimer_upgrade($oldversion) {
...
+        foreach ($records as $rec) {
+            $rec->flamedatad3 = gzcompress($rec->flamedatad3);
+            $rec->datasize = strlen($rec->flamedatad3);
+            $DB->update_record('tool_excimer_profiles', $rec);
```
keevan, line 179: Maybe use `$profiles` instead of `$records`, and similarly `$profile` instead of `$rec`.

Here's a nice resource I look at sometimes: https://github.com/jupeter/clean-code-php#use-explanatory-variables

keevan (commented):
Aside from these small changes, functionality is good once the upgrade clears, and datasize calc seems to be correct based on https://www.php.net/manual/en/function.gzencode.php

jaypha:
When it comes to 'dpwngrading' the database. the best thing is to uninstall the plugin, and reinstall it. It should never happen in production environments. I don't think Moodle has the ability to handle downgrading anyway.

db/upgrade.php
jaypha: Done

db/upgrade.php
jaypha: Done

db/install.xml
jaypha: Done

db/upgrade.php
jaypha: I appreciate the comment, but I stand by the way I have done it here.

keevan (approved):
Looks good to me

Issue #112, opened 2021-12-23
Refine the script grouping logic


brendanheywood:
* [ ] be careful with the grouping. I think we actually want some fairly nuanced logic on how we group which might mean we end up explicitly storing a 'group by' value in the db to make it easy. for example
    * [ ] most scripts we will group by just their raw script with no params
    * [ ] in select cases we will group by the full url including all params, but with redacted values so just the param keys. 
    * [ ] eg plugin files are very useful to be grouped by their component and area which are path info AND query params
    * [ ] much more advanced case which we will deal with later is cron where we will break up a single long running cron cli into multiple profiles one for each task being run. (Don't stress about this now but worth having it in the back of your mind a a use case)

jaypha:
I propose having a user setting for the scripts that are grouped with pathinfo etc. The scripts that would need this would vary from installation to installation depending on plugins installed and site usage.

PR #137, opened 2022-01-05
Issue 25 wip

### LINKED ISSUE #25: Auto break cron cli into each cron task and profile individually
We probably don’t care about the cron system itself. We care about auto profile for each task boundary. It's not immediately clear if we can cut this cleanly, but we could at minimum detect the stack trace and track it back to the cron function calls and detect if the task has changed and slice it after the fact. This means the cut point won't be perfect but it's only within the precision of a simple sample so perfectly fine.

* [ ] new set of config for minimum task length

jaypha:
resolves #25

Current concerns.
- Times such have started, finished and duration have little meaning for cron tasks, as these values cannot be adequately determined.
- A possible problem is when a timer slice falls in the middle of a task execution. Because the profiler is reset, it may result in a cron task not being recorded at all, or being recorded twice.


classes/manager.php
```diff
@@ -102,6 +102,19 @@ public static function is_profiling(): bool {
...
     }
 
+    /**
+     * True if running cron or adhoc_task scripts.
+     *
+     * @return bool
+     */
+    public static function is_cron(): bool {
+        global $SCRIPT;
+        return CLI_SCRIPT && (
```
brendanheywood, line 112: It is certainly not recommended, but there is another way to run cron from the web:

/admin/cron.php

lang/en/tool_excimer.php
```diff
@@ -107,6 +109,7 @@
 $string['scripttype_cli'] = 'CLI';
 $string['scripttype_ajax'] = 'Ajax';
 $string['scripttype_ws'] = 'Service';
+$string['scripttype_cron'] = 'Cron';
```
brendanheywood, line 112: From a naming point of view its slightly more Moodly to talk about Tasks and the Task API vs cron (I'm pretty sloppy with this myself)

settings.php
```diff
@@ -82,6 +82,16 @@
             )
         );
 
+        $settings->add(
+            new admin_setting_configtext(
+                'tool_excimer/cron_sample_threshold',
+                get_string('cron_sample_threshold', 'tool_excimer'),
+                get_string('cron_sample_threshold_desc', 'tool_excimer'),
+                '2',
```
brendanheywood, line 90: Can you convert this into a seconds based limit as the number of samples isn't easy to calculate in your head, and also when a later issue lands the sampling ratio over time may change for a given profile. The default can be reasonably large like 5 minutes

classes/cron_manager.php
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+
+    /**
+     * Extracts nodes from a flame tree that represent cron tasks.
+     *
+     * @param flamed3_node $node
+     * @return array
+     */
+    public static function extract_task_nodes(flamed3_node $node): array {
```
brendanheywood, line 87: From a quick read this looks like we are profiling the entire CLI, and then at the end we are retrospectively slicing it up into a bunch of different profiles. This might be perfectly fine, but I would have thought it was simpler to detect when a new task was running and persist the old one, then reset all the profile data and start fresh. So something to consider, happy to get push back if you think this is better

classes/cron_manager.php
```diff
@@ -0,0 +1,115 @@
...
+            $threshold = (int)get_config('tool_excimer', 'cron_sample_threshold');
+            if ($threshold < 1) {
+                $threshold = 2;
+            }
+
+            $flamed3node = flamed3_node::from_excimer($log);
+            $nodes = self::extract_task_nodes($flamed3node);
+            foreach ($nodes as $node) {
+                if ($node->value >= $threshold) {
+                    profile::save($node, $reasons, (int) $started, $duration, $current);
```
brendanheywood, line 75: When we save a task I'd like to see it's 'request' as the name of the task, and even better what task data it has but that might be too hard (if you could get this then you'd be able to split them perfectly too)

There is a lot of crossover here with this issue https://github.com/catalyst/moodle-tool_excimer/issues/112


The type of the profile is showing up are 'CLI' when I'd expect 'cron' (or better yet 'task' see below)

jaypha: Can get the name. Getting the data is too difficult.

classes/cron_manager.php
jaypha: I believe this is doable, and may be more accurate, but it would not be simpler or easier. If you want this done, I ask it be made a separate issue.

classes/cron_manager.php
```diff
@@ -0,0 +1,115 @@
...
+    protected static function process(\ExcimerLog $log, float $started): void {
+        $current = microtime(true);
+        $sampleperiod = manager::sample_period();
+        $flamed3node = flamed3_node::from_excimer($log);
+        $nodes = self::extract_task_nodes($flamed3node);
+        foreach ($nodes as $node) {
+            // Duration is between n and n+1 periods, so use n as a simple approximation.
+            $duration = $node->value * $sampleperiod;
+            $request = self::get_request($node);
+            $reasons = manager::get_reasons($request, $duration);
```
brendanheywood, line 70: Not sure best place to comment, but the max number of profiles per request is not honored. I'm happy with this being the same number for web requests, we can split this into its own config later if needed

classes/helper.php
```diff
@@ -57,6 +57,8 @@ public static function script_type_display(int $type): string {
                 return get_string('scripttype_ajax', 'tool_excimer');
             case profile::SCRIPTTYPE_WS:
                 return get_string('scripttype_ws', 'tool_excimer');
+            case profile::SCRIPTTYPE_TASK:
```
brendanheywood, line 60: I get this error on the profile.php detail page:

( ! ) Notice: Undefined offset: 0 in /var/www/moodle.local/admin/tool/excimer/classes/helper.php on line 125
--


1 | 0.0003 | 361904 | {main}( ) | .../profile.php:0
2 | 0.7364 | 17506720 | tool_excimer\helper::http_status_display( ) | .../profile.php:119

classes/cron_manager.php
brendanheywood: eg from a single cron cli run which has lots of the same task:

image

classes/profile_table.php
```diff
@@ -280,7 +280,7 @@ public function col_user(object $record): string {
     public function col_responsecode(object $record): string {
         if ($this->is_downloading()) {
             return $record->responsecode;
-        } else if ($record->scripttype == profile::SCRIPTTYPE_CLI) {
+        } else if ($record->responsecode <= 100) {
```
brendanheywood, line 283: not quite here, but for cli can you remove the redundant -

image

classes/cron_manager.php
jaypha: I did some debugging runs, and the checking code does get called. The above happens because existing profiles aren't immediately purged if you have a burst situation.

classes/helper.php
jaypha: fixed

classes/profile_table.php
jaypha: fixed

classes/cron_manager.php
brendanheywood: This doesn't sound right. At any given point we should only ever persist a new profile if it is already worse than the current ratcheting threshold. So perhaps there is something wrong with the threshold matching logic as this seems to be persisting every single task profile regardless of where.

... ah I see the problem, this all comes back to the root cause of splitting the profiles after the fact rather than resetting the profiler. I can see two other related symptoms: 1) the time started for the profile is the start of the process and not the inferred start of the task. 2) the duration of the task is the complete. It would not surprise me if there are other symptoms. 

Because of point 2, every single task profile is always worse than the previous ones. I'd still prefer it if we just stopped the profiler and reset it each time but I'll leave the internals to you.

classes/helper.php
```diff
@@ -124,6 +126,20 @@ public static function http_status_display(int $status): string {
...
     }
 
+    /**
+     * Returns status as a badge.
+     *
+     * @param int $status
+     * @return string
+     */
+    public static function status_display(int $status): string {
+        if ($status < 100) {
```
brendanheywood, line 136: I think this logic will need either the full record as input so it can look at the script type

"Exit codes are a number between 0 and 255, which is returned by any Unix command when it returns control to its parent process."

https://www.shellscript.sh/exitcodes.html

classes/task_samples.php
```diff
@@ -0,0 +1,69 @@
...
+
+namespace tool_excimer;
+
+defined('MOODLE_INTERNAL') || die();
+
+/**
+ * Class for storing samples copied over from the profiler to match a current task.
+ *
+ * @package    tool_excimer
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
```
brendanheywood, line 25: J whitespace offset :)

classes/task_samples.php
```diff
@@ -0,0 +1,69 @@
...
+ * @copyright  2021 Catalyst IT
```
brendanheywood, line 26: and its 2022!

classes/task_samples.php
```diff
@@ -0,0 +1,69 @@
...
+     * Processes the stored samples to create a profile (if eligible).
+     *
+     * @param float $finishtime
+     * @throws \dml_exception
+     */
+    public function process(float $finishtime): void {
+        $duration = $finishtime - $this->starttime;
+
+        $reasons = manager::get_reasons($this->name, $duration);
+        // TODO: move this to inside if loop before full release.
```
brendanheywood, line 59: still planning to do this comment and the one below?

slowest.php
```diff
@@ -29,7 +29,7 @@
 require_once('../../../config.php');
 require_once($CFG->libdir.'/adminlib.php');
 
-$script = optional_param('script', '', PARAM_LOCALURL);
+$script = optional_param('script', '', PARAM_RAW);
```
brendanheywood, line 32: can you add a comment here saying why we need raw over url param please

classes/helper.php
jaypha: The spanner is /admin/cron.php, which produces a HTTP response code, but is flagged as a CLI script internally. Haven't yet figured out what to do about it.

classes/task_samples.php
jaypha: doh!

classes/task_samples.php
jaypha: fixed

classes/task_samples.php
jaypha: It can always be put back in if needed.

classes/helper.php
brendanheywood: If status doesn't make sense then I'd just leave it as null. In theory it has a status but we can't get at it easily

brendanheywood: Sorry I misunderstood that and was talking about a task not the whole process. In any case it's such a small edge case I don't want to worry about it now

classes/task_samples.php
brendanheywood: I'm fine for things to land with TODO comments

slowest.php
jaypha: I have put in a comment as requested. It is need because the script name may be a task name which does not follow the URL pattern.

classes/cron_manager.php
jaypha: I have refactored this to work as you suggested. The times are still rough, +/- a sample period, but should be better than previously.

classes/helper.php
jaypha: I left a TODO note for future reference.

classes/task_samples.php
jaypha: I removed it anyway.

PR #140, opened 2022-01-06
Issue #131: Added D3 and d3-flame-graph third party libraries.

### LINKED ISSUE #131: Bundle d3 and flamegraph dependencies locally
* [ ] Add clear docs in update

https://docs.moodle.org/dev/Plugin_files#readme_moodle.txt

jaypha:
resolves #131
Created lib directory.
Added D3 library
Added d3-flame-graph library
Created lib/thirdpartylibs.xml
Updated flamegraph.mustache to use new library source.


lib/d3-flame-graph/readme_moodle.txt
```diff
@@ -0,0 +1,17 @@
+
+Home page for d3-flamegraph is https://github.com/spiermar/d3-flame-graph.
+
+You will need NPM isntalled on your system to perfrom the follwoing procedure.
```
keevan, line 4: Typo `isntalled` => `installed`

keevan: perfrom => perform

keevan: follwoing => following

lib/d3-flame-graph/readme_moodle.txt
```diff
@@ -0,0 +1,17 @@
...
+Procedure for installing d3-flame-graph
+* Create a suitable directory.
+* Perform the following commands.
+
+$ git clone https://github.com/spiermar/d3-flame-graph.git
+$ cd d3-flame-graph
+$ npm install
+
+* Copy contents of 'dist' directory to lib/d3-flamegraph in plugin.
+* Copy LICENSE to lib/d3-flamegraph in plugin.
```
keevan, line 15: I wonder if the whole thing could be commands we can use to do a quick update.
```suggestion
* Copy contents of 'dist' and license directory to lib/d3-flamegraph in plugin.
cp -r somewhere/dist somewhere/LICENSE -t lib/d3-flame-graph
```

keevan: Also in your example the target folder has a different name from the one you've provide in this commit

lib/d3-flamegraph => lib/d3-flame-graph

lib/d3/readme_moodle.txt
```diff
@@ -0,0 +1,7 @@
+
+Procedure for including D3 library into Moodle.
+
+1. The download website is https://d3js.org.
+2. Download the latest version in a tgz file via a link given on the above page.
+3. Extract from the tgz file and move the contents of the 'package/dist' directory and LICENSE into lib/d3 of the plugin.
+
```
keevan, line 7: Similarly to the previous point regarding d3-flame-graph, could you make this into a list of commands and the steps as command comments?

An example URL would be nice to go through the update process as easily as possible without needing to do much manual work.

lib/d3/readme_moodle.txt
```diff
@@ -0,0 +1,7 @@
...
```
keevan, line 1: Remove the first empty line as I don't think it's needed.

lib/d3-flame-graph/readme_moodle.txt
```diff
@@ -0,0 +1,17 @@
+
```
keevan, line 1: Remove the first empty line as I don't think it's needed.

keevan (commented):
The rest is probably fine but I'll need to re-test once the instructions are updated.

Thanks Jason

lib/thirdpartylibs.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0"?>
+<libraries>
```
brendanheywood, line 2: this needs to be up a dir level

/thirdpartylibs.xml

templates/flamegraph.mustache
```diff
@@ -172,8 +172,8 @@
 <div id="details" style="min-height: 1.5rem; clear: both;">
 </div>
 
-<script type="text/javascript" src="https://d3js.org/d3.v4.min.js"></script>
-<script type="text/javascript" src="https://cdn.jsdelivr.net/npm/d3-flame-graph@4.0.6/dist/d3-flamegraph.min.js"></script>
+<script type="text/javascript" src="/admin/tool/excimer/lib/d3/d3.min.js"></script>
```
brendanheywood, line 175: It's better to route the serving of these through /theme/javascript.php which will add all the proper cache busting params and caching headers to it gets cached by cdn / varnish

lib/d3-flame-graph/readme_moodle.txt
brendanheywood: also worth having a step to fully remove the dir contents first so we aren't merging files across versions

lib/thirdpartylibs.xml
brendanheywood: To know if this is working, look at this admin page:

/admin/thirdpartylibs.php

templates/flamegraph.mustache
jaypha: How does this work?

PR #144, opened 2022-01-11
Issue #112: Refine grouping of profiles.

### LINKED ISSUE #112: Refine the script grouping logic

jaypha:
resolves #112

Created context class.
Moved environment and context related functions to context.
Added get_group_by_value() to context.
Added groupby column to database.
Slowest - grouped page now groups using the groupby value.


classes/context.php
```diff
@@ -0,0 +1,218 @@
...
+
+    /**
+     * Gets a value to be used for grouping profiles.
+     *
+     * @param string $request
+     * @param string $pathinfo
+     * @param string $params
+     * @return string
+     */
+    public static function get_groupby_value(string $request, string $pathinfo, string $params): string {
```
brendanheywood, line 131: This is the meat of the whole patch, it would be good to see see unit tests around it with a dataProvider

There logic here is going to end up having a bunch of logic depends on the scripts. In most scripts we will only use the script name, but in some we will use the script + the params, but we will redact the values of the params. In other cases we will redact values in the query path too.

This isn't quite what we want but its what I've used when do some log analysis which is vague along the same lines.

https://github.com/brendanheywood/scripts/blob/master/log-redact

Mostly we just need a bunch of examples to put into the dataProvider to flesh it out, one example:

/pluginfile.php/1/core_admin/logocompact/300x300/1641856751/moodle-icon.jpg

should be grouped as:

/pluginfile.php/x/core_admin/logocompact/xxx

/pluginfile.php/7282414/mod_book/chapter/644676/electrophoresis%202.PNG

as

pluginfile.php/x/mod_book/chapter/xxx

PR #148, opened 2022-01-13
Issue 146 wip

### LINKED ISSUE #85: Refactor: convert profile to a persistent
This class is already very close to the shape of a persistent so seems a natural fit

https://docs.moodle.org/dev/Persistent

### LINKED ISSUE #146: Make profile more object oriented.
There is a lot of areas where using an object for a profile before saving would make things easier. For example, a lot of parameters get passed to profile::save(). It would be easier and neater to have a object to do this.

This would be a precursor to issue #85. It would have persistence in mind, but wouldn't go all the way yet.

- [ ] Move out of profile, all functions that are not directly relevant to profiles, particularly functions that glean information from the environment. Move them to either manager, or a new class.

- [ ] Make profile into an object and add functions to add/manipulate data.

- [ ] Convert save() into a non-static.

jaypha:
resolves #146
resolves #85
Refactoring to make profile more object oriented.


classes/context.php
```diff
@@ -0,0 +1,192 @@
...
+
+/**
+ * Functions that extract information from the execution environment.
+ *
+ * @package   tool_excimer
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>, Kevin Pham <kevinpham@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class context {
```
brendanheywood, line 29: is there a better name so there is no confusion with moodle contexts?

classes/manager.php
```diff
@@ -16,13 +16,11 @@
...
 namespace tool_excimer;
 
-defined('MOODLE_INTERNAL') || die();
-
 /**
  * Primary controller class for handling Excimer profiling.
  *
  * @package   tool_excimer
- * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>, Kevin Pham <kevinpham@catalyst-au.net>
```
brendanheywood, line 23: super minor, but I think this works better generally as 2 `@author` lines

classes/profile.php
```diff
@@ -16,35 +16,41 @@
...
 
-    /** Report section - slowest - lists the slowest profiles first */
-    const REPORT_SECTION_SLOWEST = 'slowest';
+    /** Reason - FLAMEALL - Toggles profiling for all subsequent pages, until FLAMEALLSTOP param is passed as a page param. */
+    const REASON_FLAMEALL = 0b0100;
 
-    /** Report section - unfinished - lists profiles of scripts that did not finish */
-    const REPORT_SECTION_UNFINISHED = 'unfinished';
+    /** Reason - NONE - Default fallback reason value, this will not be stored. */
+    const REASON_NONE = 0b0000;
```
brendanheywood, line 41: Can you put these in order so 0 would be first

classes/profile.php
```diff
@@ -16,35 +16,41 @@
...
  * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 class profile {
+    const TABLE = 'tool_excimer_profiles';
+
 
-    /** Request's fallback value for when the $SCRIPT is null */
-    const REQUEST_UNKNOWN = 'UNKNOWN';
+    /** Reason - MANUAL - Profiles are manually stored for the request using FLAMEME as a page param. */
+    const REASON_MANUAL   = 0b0001;
```
brendanheywood, line 32: kinda off topic but it also seems inconsistent now in hindsight that we have REASON_FLAMEALL vs REASON_MANUAL. I think the latter should be REASON_FLAMEME

classes/profile.php
```diff
@@ -53,160 +59,119 @@ class profile {
...
-        return $parameters;
+    // TODO: try to find a way to eliminate the need for this function.
+    public static function get_num_profiles(): int {
+        global $DB;
+        return $DB->count_records(self::TABLE, []);
     }
 
     /**
-     * Gets a single profile, including data.
+     * Loads a profile from disk.
```
brendanheywood, line 72: from db, not disk

classes/context.php
jaypha: I find that one of the hardest problems in programming is coming up with decent names. Is 'environment' OK?

brendanheywood: Can't agree more.

These are all various types of script or request metadata, so how about something like request_metadata or script_metadata

jaypha: script_metadata it is.

classes/manager.php
jaypha: done

classes/profile.php
jaypha: The function has been erased as redundant.

classes/profile.php
jaypha: done

classes/profile.php
jaypha: done

PR #149, opened 2022-01-14
Issue 112a wip

### LINKED ISSUE #112: Refine the script grouping logic

jaypha:
resolves #112 
Decide to create a new PR to keep things neat.

### DIFF
--- a/classes/profile.php
+++ b/classes/profile.php
@@ -77,7 +77,7 @@ class profile extends persistent {
      *
      * @return profile
      */
-    public static function get_running_profile(?string $request = null): profile {
+    public static function get_running_profile(): profile {
         if (!isset(self::$runningprofile)) {
             self::$runningprofile = new profile();
         }
--- a/classes/profile.php
+++ b/classes/profile.php
@@ -77,7 +77,7 @@ class profile extends persistent {
      *
      * @return profile
      */
-    public static function get_running_profile(): profile {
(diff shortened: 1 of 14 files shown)


classes/profile.php
```diff
@@ -379,32 +380,33 @@ protected static function define_properties(): array {
         return [
             'reason' => ['type' => PARAM_INT],
             'scripttype' => ['type' => PARAM_INT],
-            'method' => ['type' => PARAM_ALPHA],
-            'created' => ['type' => PARAM_INT],
+            'method' => ['type' => PARAM_ALPHA, 'default' => ''],
```
brendanheywood, line 383: will ci complain if we index these all the line up in columns?

jaypha: It does.

PR #150, opened 2022-01-16
Issue #19: Added privacy classes

### LINKED ISSUE #19: Add privacy classes
This thing will store potentially tracking data on any individual and see what they did. No different to say the moodle logs or server access logs but should be declared

jaypha:
resolves #19

### DIFF
--- a/classes/privacy/provider.php
+++ b/classes/privacy/provider.php
@@ -61,7 +61,6 @@ class provider implements \core_privacy\local\metadata\provider,
      */
     public static function get_contexts_for_userid(int $userid) : contextlist {
         $contextlist = new contextlist();
-        $contextlist->add_user_context($userid);
         $contextlist->add_system_context();
         return $contextlist;
     }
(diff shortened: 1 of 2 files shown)


classes/privacy/provider.php
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Get the list of contexts that contain user information for the specified user.
+     *
+     * @param   int         $userid     The user to search.
+     * @return  contextlist   $contextlist  The contextlist containing the list of contexts used in this plugin.
+     */
+    public static function get_contexts_for_userid(int $userid) : contextlist {
+        $contextlist = new contextlist();
+        $contextlist->add_user_context($userid);
```
Peterburnett, line 64: What lives in the user context? This is typically data that is viewable by the user, such as profile data. I expected this to be all system context

jaypha: The line has been deleted.

PR #155, opened 2022-01-20
Issue #24: Double sampling period.

### LINKED ISSUE #24: Bonus points: Compress extremely large profiles so they have a fixed upper limit
Lets say we have an extremely long profile, maybe an integration web service or something that runs on the cli and a very long report which hopefully streams. The larger the profile is, the less we care about the fine details. The bottlebeck will be the same either way. Let’s say we record cron or a report that take 4 hours.

So the idea is that we can arbitraily set some maximum size for the profile itself, and if larger profiles come in the we can reduce the sampling period. I can see two ways this might happen:

1) can we tell excimer to change the sampling frequency after it has started running? So for instance after 1 minute we double the sample rate. After 2 minutes we double again. After 4 minutes we double.

2) Even if we can't, we could just selectively ignore some of the callbacks.

3) After the fact, before we persist the profile, can we do some time slicing and reduce the resolution of the profile data?

Question: If a profile has variable sized time slices with the flame graph just work because it’s know each time slice?

jaypha:
resolves #24 
Added setting 'half life'.
Added code to have the setting double roughly in accordance with half-life setting.

### DIFF
--- a/classes/cron_manager.php
+++ b/classes/cron_manager.php
@@ -31,20 +31,18 @@ class cron_manager {
     /**
      * Sets callbacks to handle cron profiling.
      *
-     * @param \ExcimerProfiler $profiler
-     * @param \ExcimerTimer $timer
-     * @param float $started
+     * @param object $profiler
      * @throws \dml_exception
      */
-    public static function set_callbacks(\ExcimerProfiler $profiler, \ExcimerTimer $timer): void {
-        $timer->setCallback(function() use ($profiler) {
-            cron_manager::on_interval($profiler);
+    public static function set_callbacks(object $profiler): void {
+        $profiler->timer->setCallback(function() use ($profiler) {
+            cron_manager::on_interval($profiler->profiler);
         });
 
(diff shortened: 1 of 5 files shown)


jaypha:
I haven't figured out how to stretch out the nodes. The plugin does increase the sampleperiod in accordance with the half-life settings, but the graph displays only hits with no time information.

jaypha:
Copied from the issue.

I have hand tested the half-life sampling and believe it to be working. The sampling period gets progressively longer each time the partial save period is triggered. I have seen it working even when the sampling period exceeds the partial save period. It remains to be seen how it would work on a truly long running profile.

I have not been able to figure out how to make a unit test for it.

This has not been implemented yet for cron tasks. They still have flat timing.

As I mentioned above, flame graph data does not contain time information. Each node contains only a name and a number. For excimer, it is the name of the function and the number of times it has been called. It is not even in chronological order.
Therefore the flame data graph cannot 'stretch' as the time interval changes. It has no knowledge of such.

PR #162, opened 2022-02-15
Issue 161 wip

### LINKED ISSUE #161: Bug when running cron
No idea what this is about just hitting doing unrelated stuff

$ php admin/cli/adhoc_task.php -e
Server Time: Tue, 08 Feb 2022 15:41:28 +1100

Execute adhoc task: tool_testtasks\task\timed_adhoc_task
++ on_interval called with no profile ++
* line 73 of /admin/tool/excimer/classes/cron_manager.php: call to debugging()
* line 41 of /admin/tool/excimer/classes/cron_manager.php: call to tool_excimer\cron_manager::on_interval()
* line 689 of /lib/classes/date.php: call to tool_excimer\cron_manager::tool_excimer\{closure}()
* line 129 of /lib/classes/date.php: call to core_date::init_zones()
* line 272 of /lib/classes/date.php: call to core_date::normalise_timezone()
* line 2350 of /lib/moodlelib.php: call to core_date::get_user_timezone()
* line 327 of /calendar/type/gregorian/classes/structure.php: call to date_format_string()
* line 2292 of /lib/moodlelib.php: call to calendartype_gregorian\structure->timestamp_to_date_string()
* line 390 of /lib/cronlib.php: call to userdate()
* line 303 of /lib/cronlib.php: call to cron_set_process_title()
* line 198 of /lib/cronlib.php: call to cron_run_inner_adhoc_task()
* line 131 of /admin/cli/adhoc_task.php: call to cron_run_adhoc_tasks()
... started 15:41:28. Current memory use 18.6 MB.

jaypha:
Resolves #161

Introduce a small timer period to make the timer and profiler trigger separately for cron tasks.
Use a loop to examine log entries on each interval.

There may still be incidents of 'on_interval called with no profile'. This is not because of any flaw in the logic, but because the timers are volatile on starting. The first action does not line up with the sample period set. The actions become more stable after the first one.

This is not known to affect profiling of cron tasks.

### DIFF
--- a/classes/cron_manager.php
+++ b/classes/cron_manager.php
@@ -31,20 +31,23 @@ class cron_manager {
     /**
      * Sets callbacks to handle cron profiling.
      *
-     * @param \ExcimerProfiler $profiler
-     * @param \ExcimerTimer $timer
-     * @param float $started
+     * @param object $profiler
      * @throws \dml_exception
      */
-    public static function set_callbacks(\ExcimerProfiler $profiler, \ExcimerTimer $timer): void {
-        $timer->setCallback(function() use ($profiler) {
-            cron_manager::on_interval($profiler);
+    public static function set_callbacks(object $profiler): void {
+        $profiler->timer->setCallback(function() use ($profiler) {
+            $profiler->timer->setPeriod($profiler->sampleperiod);
+            $profiler->timer->setCallback(function() use ($profiler) {
+                cron_manager::on_interval($profiler);
+            });
+            $profiler->timer->stop();
+            $profiler->timer->start();
         });
 
         \core_shutdown_manager::register_function(
-            function() use ($profiler, $timer) {
-                $timer->stop();
-                $profiler->stop();
+            function() use ($profiler) {
+                $profiler->timer->stop();
+                $profiler->profiler->stop();
                 if (self::$currenttask) {
                     self::$currenttask->process(microtime(true));
                 }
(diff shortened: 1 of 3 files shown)


jaypha:
With the introduction of a loop, it is no longer needed to make the two timers line up. So the cron ExcimerTimer now uses the partial save interval like the regular manager. This not only simplifies things in general, but fixes this issue completely.

classes/cron_manager.php
```diff
@@ -63,39 +62,29 @@ function() use ($profiler, $timer) {
...
-        }
-        $sample = $log->current();
-        $log->next();
-        if ($log->valid()) {
-            debugging('extra profile sample found in on_interval');
-        }
-
-        $taskname = self::findtaskname($sample);
-        $currenttime = microtime(true);
+        foreach ($log as $sample) {
```
Peterburnett, line 70: It looks like this would significantly increase the computation time on every interval, and compounding. Would it be possible to store the index previously accessed, and then resume from that point in log entries (assuming the history of the log is immutable)

classes/cron_manager.php
```diff
@@ -63,39 +62,29 @@ function() use ($profiler, $timer) {
...
-        if ($taskname && (self::$currenttask == null)) {
-            self::$currenttask = new task_samples($taskname, self::$sampletime);
-        }
+            if (self::$currenttask) {
+                self::$currenttask->add_sample($sample);
+            }
 
-        if (self::$currenttask) {
-            self::$currenttask->add_sample($sample);
+            self::$sampletime = $currenttime;
```
Peterburnett, line 86: It looks like this isnt necessary inside a loop, instead it could be set outside the loop, based on the context of the most recent log sample?

classes/manager.php
```diff
@@ -118,58 +120,55 @@ public static function sample_period(): float {
      * @throws \dml_exception
      */
     public static function init(): void {
-        $sampleperiod = self::sample_period();
+        self::$profiler = new \stdClass();
```
Peterburnett, line 123: Why did these need to become static. Could this lead to any weird collisions with concurrent single process profiling?

classes/cron_manager.php
Peterburnett: I also think this should be unit tested with the conditions that were causing the problem

jaypha: Calling 'flush()' empties the profiler of its logs, so there is no re-processing.

jaypha:
This PR is not being merged, PR #163 was merged instead.

PR #163, opened 2022-02-16
Issue #161: Fix "on_interval called with no profile" bug

jaypha:
Refactored cron_manager::on_interval. ExcimerTimer no longer needs to trigger
once per sample.


classes/cron_manager.php
```diff
@@ -64,38 +66,32 @@ function() use ($profiler, $timer) {
...
-        if (!$log->valid()) {
-            debugging('on_interval called with no profile');
-            return;
-        }
-        $sample = $log->current();
-        $log->next();
-        if ($log->valid()) {
-            debugging('extra profile sample found in on_interval');
-        }
+        foreach ($log as $sample) {
```
Peterburnett, line 74: Same thought here regarding performance. For a long running script this could get very heavy. I think its worth storing the count of samples processed, and then chopping off the first N samples that we know are already processed. I think the only additional data to track would be the number of samples, plus the ending task name.

That will reduce this function from n time to constant time, which will keep the footprint low here

jaypha: Calling flush() on ExcimerProfile removes all logs on the profiler, so next time it comes around, it is all new samples. No sample is processed twice.

jaypha: Not sure why we need to track the ending task name.

Peterburnett: That would only have been to keep the task list state intact if we were picking up in a new loop, Mightve been unnecessary anyway. In the case that its flushed, my comment is already implemented :)

PR #165, opened 2022-02-17
Refactor: use PHPUnit stubs

jaypha:
Changed tool_excimer_flamed3_node_test to use a PHPUnit stub rather that a
custom mock class.

### DIFF
--- a/tests/excimer_mockery.php
+++ b/tests/excimer_mockery.php
@@ -1,0 +1,154 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_excimer;
+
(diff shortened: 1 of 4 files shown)


jaypha:
The stubs for ExcimerProfiler and ExcimerLog are not used as yet. But they will be used in future.

tests/excimer_mockery.php
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * Intermediary class to provide stubs for Excimer classes.
+ *
+ * @package    tool_eximer
+ * @author     Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright  2022, Catalyst IT
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class excimer_mockery extends \advanced_testcase {
```
Peterburnett, line 29: Because this ends up getting used as a testcase, I think this should likely be excimer_testcase

tests/excimer_mockery.php
```diff
@@ -0,0 +1,154 @@
...
+            $logentries[] = $this->get_log_entry_stub($tracedef);
+        }
+        $logentries = new \ArrayObject($logentries);
+        $iterator = $logentries->getIterator();
+
+        $stub = $this->getMockBuilder(\ExcimerLog::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $stub->method('current')
```
Peterburnett, line 91: These feel dangerously close to doing a foreach on iterator method names and logentries method names.

I think it likely should go that way. 2 arrays, on for the iterator functions and one for the logentries functions.
then 

method($string)->willReturnCallback(function() use ($iterator, $string){ return $iterator->$string()});

tests/excimer_mockery.php
jaypha: Awww. You don't like my mockery pun?

jaypha: I changed the name.

PR #166, opened 2022-02-23
Issue #164: Refactoring for more object orientedness.

### LINKED ISSUE #164: Refactor to improve flexibility
- Replace largely static classes with objects.
- Have a manager class that accepts a 'processor' object to handle the processing of samples.
- Create two processor classes, for cron and regular profiling.
- Change internal time variables to store values as seconds.

jaypha:
Closes #164 

Add new classes cron_processor and regular_processor to process samples for cron runs and regular runs respectively.
Refactor manager class to be an object that takes a processor.
Move get_timer_interval() and get_sampling_period() to script_metadata.
Move get_min_duration_for_group_and_reason(), get_min_duration_for_reason() and clear_min_duration_cache_for_reason() to profile.
Refactor all internal time storing variables to store as seconds.

cron_processor replaces cron_manager, and regular_processor takes over many tasks of manager.
sample_set replaces task_samples.


classes/cron_processor.php
```diff
@@ -0,0 +1,136 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_excimer;
+
+class cron_processor implements processor {
+    public $sampletime;
+    public $currenttask = null;
+
+    public function init(manager $manager): void {
```
Peterburnett, line 23: Moodle coding standards leave :void out and only declare the return type for data returns

classes/cron_processor.php
```diff
@@ -0,0 +1,136 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_excimer;
+
+class cron_processor implements processor {
+    public $sampletime;
```
Peterburnett, line 20: PHPDoc typehinting these would be nice

classes/cron_processor.php
```diff
@@ -0,0 +1,136 @@
...
+                $manager->get_timer()->stop();
+                $manager->get_profiler()->stop();
+                $this->on_interval($manager);
+                if ($this->currenttask) {
+                    $this->process($manager, microtime(true));
+                }
+            }
+        );
+    }
+
```
Peterburnett, line 41: Double whitespace here

classes/cron_processor.php
Peterburnett, line 23: Also missing PHPDoc for this class

classes/manager.php
```diff
@@ -31,8 +31,64 @@ class manager {
...
+     * Initialises the manager.
+     *
+     * @param processor $processor
+     * @throws \coding_exception
+     */
+    public function __construct(processor $processor) {
+        $this->processor = $processor;
+    }
+
+    public function init() {
```
Peterburnett, line 61: Missing PHPDoc block for function

classes/cron_processor.php
```diff
@@ -0,0 +1,136 @@
...
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_excimer;
+
```
keevan, line 18: Missing https://docs.moodle.org/dev/Coding_style#Class_declarations

You can ignore the file declaration if you have one class and have a class declaration.

classes/manager.php
```diff
@@ -304,69 +185,33 @@ public static function get_min_duration_for_reason(int $reason, bool $usecache =
...
-        if ($durationms <= self::min_duration()) {
+        if ($duration <= $this->processor->get_min_duration()) {
             return false;
         }
 
         // If a min duration exists, it means the quota is filled, and only
         // profiles slower than the fastest stored profile should be stored.
-        $minduration = self::get_min_duration_for_reason(profile::REASON_SLOW);
-        if ($minduration && $durationms <= $minduration) {
+        $minduration = profile::get_min_duration_for_reason(profile::REASON_SLOW);
```
Peterburnett, line 199: Should this be a static reference? I can imagine a world where the "slow" duration varies between profile objects, and it seems like being static here is against the pattern of the rest of the commits

classes/profile.php
```diff
@@ -196,13 +185,121 @@ public function save_record(): int {
...
+     * Cost: 1 cache read (ideally)
+     * Otherwise: 1 cache read, 1 DB read and 1 cache write.
+     *
+     * @param  string $request holds the request url
+     * @param  int $reason the profile type or REASON_*
+     * @param  bool $usecache whether or not to even bother with caching. This allows for a forceful cache update.
+     *
+     * @return float duration (as seconds) of the fastest profile for a given reason and request/page.
+     */
+    public static function get_min_duration_for_group_and_reason(string $group, int $reason, bool $usecache = true): float {
```
Peterburnett, line 207: Follow on from the comments above around static access. Feels weird to have these be static methods when it could be object methods. I suspect that you will always have a profile object anyway when needing to access these methods, but if not, perhaps it could be refactored to a helper static class?

classes/regular_processor.php
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Processor for regular profiling.
+ *
+ * @package   tool_excimer
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class regular_processor implements processor {
```
Peterburnett, line 27: Instead of regular_processor, perhaps web_processor?

classes/cron_processor.php
jaypha: done

classes/cron_processor.php
jaypha: done

classes/cron_processor.php
jaypha: done

classes/cron_processor.php
jaypha: done

classes/manager.php
jaypha: done

classes/cron_processor.php
jaypha: done

classes/regular_processor.php
jaypha: done

classes/manager.php
jaypha: This comment, and the next, address the same thing. So this response is for both. These functions were written by Kevin, so I avoid messing with them unless I have to. I moved them to profile because they are relevant to profiles (they access the profile database table). They are static because are aggregate functions. They do not refer to a specific item.

PR #168, opened 2022-02-25
Limit the number of samples recorded.

### LINKED ISSUE #24: Bonus points: Compress extremely large profiles so they have a fixed upper limit

jaypha:
Resolves #24 
Adds a filtering mechanism to limit the number of samples that get saved.
Whenever the number of samples reaches N, the filtering rate is doubled and every second sample stored is removed.
With the filtering rate, only every Rth sample gets stored.


classes/sample_set.php
```diff
@@ -30,18 +30,53 @@ class sample_set {
...
+    public $current = 0;
+
+    /**
+     * Constructs the sample set.
+     *
+     * @param string $name
+     * @param float $starttime
+     * @param float|null $limit
+     */
+    public function __construct(string $name, float $starttime, int $limit = null) {
```
keevan, line 45: Curious when `$limit` might be null. I haven't read through the rest yet at the time of writing this, but seems like some assumptions made that it is set.

classes/sample_set.php
```diff
@@ -30,18 +30,53 @@ class sample_set {
     public $samples = [];
     public $starttime;
 
+    public $limit;
 
-    public function __construct($name, $starttime) {
+    public $filterrate = 1;
+    public $current = 0;
```
keevan, line 36: A description here would be great. "Current..." what? Is this meant to be a counter / index / incremental id?

classes/sample_set.php
```diff
@@ -30,18 +30,53 @@ class sample_set {
...
      *
      * @param array $sample
      */
     public function add_sample(\ExcimerLogEntry $sample): void {
-        $this->samples[] = $sample;
+        if (count($this->samples) === $this->limit) {
+            $this->apply_doubling();
+        }
+        $this->current += 1;
+        if ($this->current == $this->filterrate) {
```
keevan, line 61: Should use `===` instead of `==` because it seems it should always be an `int` comparison.

classes/script_metadata.php
```diff
@@ -316,4 +316,15 @@ public static function get_sampling_period(): float {
...
+    /**
+     * Returns the sampling double rate.
+     *
+     * @return int
+     * @throws \dml_exception
+     */
+    public static function get_sampling_doublerate(): int {
+        $rate = (int)get_config('tool_excimer', 'doublerate');
+        $insensiblerange = $rate >= 0;
+        return $insensiblerange ? $rate : 1024;
```
keevan, line 328: Was a bit hard to read, please consider the following

```suggestion
        if ($rate <= 0) {
            return 1024;
        }
        return $rate;
```

classes/script_metadata.php
```diff
@@ -316,4 +316,15 @@ public static function get_sampling_period(): float {
         return round($insensiblerange ? $period : self::SAMPLING_PERIOD_DEFAUILT, 3);
     }
 
+    /**
+     * Returns the sampling double rate.
```
keevan, line 320: A longer description here would be useful. What is this value, and what is it being used for? I see this "rate value" is being put into a limit field for the sample_set, but not sure how the values translate across.

classes/web_processor.php
```diff
@@ -33,15 +33,29 @@ class web_processor implements processor {
      */
     protected $profile;
 
+    /**
+     * @var sample_set
+     */
```
keevan, line 38:
```suggestion
    /** @var sample_set */
```
Can be a one liner, same for similar class props

keevan (commented):
I've read up to `cli/simtask.php` for now, but will continue this next week.

tests/tool_excimer_sample_set_test.php
```diff
@@ -0,0 +1,127 @@
...
+        $entries = [
+            $this->get_log_entry_stub(['a']),
+            $this->get_log_entry_stub(['b']),
+            $this->get_log_entry_stub(['c']),
+        ];
+
+        $set = new sample_set('a', 0);
+
+        foreach ($entries as $entry) {
+            $set->add_sample(($entry));
```
keevan, line 40: Unnecessary double parenthesis

tests/tool_excimer_sample_set_test.php
```diff
@@ -0,0 +1,127 @@
...
+            $this->get_log_entry_stub(['b']),
+            $this->get_log_entry_stub(['c']),
+            $this->get_log_entry_stub(['d']),
+        ];
+
+        $set = new sample_set('a', 0);
+        $set->apply_doubling();
+
+        foreach ($entries as $entry) {
+            $set->add_sample(($entry));
```
keevan, line 58: Unnecessary double parenthesis

tests/tool_excimer_sample_set_test.php
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        $this->assertEquals([$entries[1], $entries[3]], $set->samples);
+
+        $set = new sample_set('a', 0);
+        $set->apply_doubling();
+        $set->apply_doubling();
```
keevan, line 65: Comments here would be good to know what the sampling interval should be at currently, or what the expected behaviour should be for anything coming after.

tests/tool_excimer_sample_set_test.php
```diff
@@ -0,0 +1,127 @@
...
+
+        foreach ($entries as $entry) {
+            $set->add_sample(($entry));
```
keevan, line 68: Unnecessary double parenthesis

tests/tool_excimer_sample_set_test.php
```diff
@@ -0,0 +1,127 @@
...
+        ];
+
+        $set = new sample_set('a', 0);
+
+        foreach ($entries as $entry) {
+            $set->add_sample(($entry));
+        }
+
+        $set->apply_doubling();
+        $this->assertEquals([$entries[1], $entries[3]], $set->samples);
```
keevan, line 89: Same as previous, I think it's easier to read if broken up to separate lines. At a glance it is very easy to misinterpret the 2 array entries as different params for the assertEquals function.

tests/tool_excimer_sample_set_test.php
```diff
@@ -0,0 +1,127 @@
...
+            $set->add_sample(($entry));
+        }
+
+        $this->assertEquals([$entries1[1], $entries1[3], $entries1[5]], $set->samples);
+
+        foreach ($entries2 as $entry) {
+            $set->add_sample(($entry));
+        }
+
+        $this->assertEquals([$entries1[3], $entries2[1], $entries2[5]], $set->samples);
```
keevan, line 125: Not being sure what's tested here aside from the fact it's testing some sort of doubling (function name).

I assume (after reading all the code) it's something to do with the limit set to 4, but at a glance I'm unsure how it's testing this or what the expectations should be. It would greatly help if there were comments here.

tests/tool_excimer_script_metadata_test.php
```diff
@@ -26,6 +26,14 @@
  */
 class tool_excimer_script_metadata_test extends \advanced_testcase {
 
+    /**
+     * Set up before each test
+     */
+    protected function setUp(): void {
```
keevan, line 32: You can remove the void typehint here

tests/tool_excimer_script_metadata_test.php
```diff
@@ -26,6 +26,14 @@
...
+        parent::setUp();
+        $this->resetAfterTest();
+    }
+
     public function test_stripparamters(): void {
```
keevan, line 37: > You can remove the void typehint here

Same here.

classes/sample_set.php
jaypha: It was my original idea to be able to switch it off by setting it to null, but now it has been decided to always set it to a value.

classes/sample_set.php
jaypha: done

classes/web_processor.php
```diff
@@ -28,20 +28,27 @@
...
+            script_metadata::get_samplelimit()
+        );
+
         $this->profile = new profile();
-        $this->profile->add_env(script_metadata::get_request());
-        $this->profile->set('created', (int) $manager->get_starttime());
+        $this->profile->add_env($this->sampleset->name);
+        $this->profile->set('created', $this->sampleset->starttime);
 
         $manager->get_timer()->setCallback(function($s) use ($manager) {
```
keevan, line 53: Not exactly in this patch but since it's in view of the diff, can you please change the anonymous function to have a space after the function keyword please? This makes it consistent with the PSR standards (which Moodle defers to when not specified directly)

https://www.php-fig.org/psr/psr-12/#47-method-and-function-calls

classes/web_processor.php
```diff
@@ -73,14 +80,17 @@ public function get_min_duration(): float {
      * @throws \dml_exception
      */
     public function process(manager $manager, bool $isfinal): void {
-        $log = $manager->get_profiler()->getLog();
+        $log = $manager->get_profiler()->flush();
+        foreach ($log as $entry) {
+            $this->sampleset->add_sample($entry);
+        }
```
keevan, line 86: Looking at this and the tests, I think it would be good to have a `add_many_samples` function or similar, to do handle this loop consistently all the time. This would also reduce the logic done in general (since flush and samples need to always be added like this) and also would simplify the handling in tests.

tests/tool_excimer_sample_set_test.php
```diff
@@ -0,0 +1,183 @@
...
+
+        // Only every 2nd sample should be recorded in sample set.
+        $this->assertEquals($expected1, $set->samples);
+
+        $set = new sample_set('a', 0, 1024);
+        $set->apply_doubling();
+        $set->apply_doubling();
+
+        // Filter rate should be 4, thus, only every 4th sample should be recorded in sample set.
+
```
keevan, line 88: Please kill this empty line.

tests/tool_excimer_sample_set_test.php
```diff
@@ -0,0 +1,183 @@
...
+        $set = new sample_set('a', 0, 1024);
+
+        foreach ($samples as $sample) {
+            $set->add_sample($sample);
+        }
+
+        $set->apply_doubling();
+
+        // Half of the samples should be stripped (every 2nd).
+        $this->assertEquals($expected1, $set->samples);
```
keevan, line 125: Maybe this format would read a bit better? (What you intend to do and what the expected behaviours are)
```suggestion
        // Every 2nd sample should be stripped after doubling.
        $set->apply_doubling();
        $this->assertEquals($expected1, $set->samples);
```

tests/tool_excimer_sample_set_test.php
```diff
@@ -0,0 +1,183 @@
...
+        foreach ($samples1 as $sample) {
+            $set->add_sample($sample);
+        }
+
+        // By this point apply_doubling should have been invoked once.
+        $this->assertEquals($expected1, $set->samples);
+
+        foreach ($samples2 as $sample) {
+            $set->add_sample($sample);
+        }
```
keevan, line 178: Same as previous, just marking it. (`add_many_samples`)

tests/tool_excimer_script_metadata_test.php
```diff
@@ -26,7 +26,15 @@
...
-    public function test_stripparamters(): void {
+    /**
+     * Set up before each test
+     */
+    protected function setUp(): void {
+        parent::setUp();
+        $this->resetAfterTest();
+    }
+
+    public function test_stripparamters() {
```
keevan, line 37: Misspelling here. Also I think as previous comments mentioned, it would read better as `test_strip_parameters`

keevan (changes requested):
Hi @jaypha, I've looked over this PR again and I think a few more changes would be better.

tests/tool_excimer_sample_set_test.php
```diff
@@ -0,0 +1,167 @@
...
+
+        // Half of the samples should be stripped again, leaving every 4th from the original.
+        $set->apply_doubling();
+        $this->assertEquals($expected2, $set->samples);
+    }
+
+    /**
+     * Tests the invoking of apply_doubling from within add_sample.
+     */
+    public function test_doubling() {
```
keevan, line 124: Can you change the name of this to something like `test_automatic_doubling_when_adding_samples` or similar? The test name _should_ describe what it's testing. The body of this test seems to indicate the suggested new name, whereas the current name seems more focused on testing just the apply_doubling method which is somewhat covered in other tests already.

tests/tool_excimer_sample_set_test.php
```diff
@@ -0,0 +1,167 @@
...
+            $this->get_log_entry_stub(['j']),
+            $this->get_log_entry_stub(['k']),
+            $this->get_log_entry_stub(['l']),
+        ];
+        // This is every 4th element of $sample1 + $sample2.
+        $expected2 = [
+            $samples1[3],
+            $samples2[1],
+            $samples2[5]
+        ];
```
keevan, line 153: > This has the same effect as doubling the sampling period

Affected by previous comment in the doubling function.

classes/sample_set.php
```diff
@@ -17,31 +17,85 @@
...
+    public function add_sample(\ExcimerLogEntry $sample) {
+        if (count($this->samples) === $this->samplelimit) {
+            $this->apply_doubling();
+        }
+        $this->counter += 1;
+        if ($this->counter === $this->filterrate) {
+            $this->samples[] = $sample;
+            $this->counter = 0;
+        }
+    }
```
keevan, line 68: I'm not sure if it's exactly `1:1` with the `samplelimit_desc`, it does not have to be but it's a small detail that I'm not sure if it's worth fixing. It seems what is happening here is, the limit is reached, you are halving what you have, and you are adding a sample to the list. But this extra sample, would have been an entry from the _previous_ sample rate. To be more _correct_ do you think it should instead be returning early whenever the sample limit has been met?

keevan (commented):
There are a few other review items shown that have not been commented on or fixed up, if you could please attend to them as well.

classes/sample_set.php
```diff
@@ -17,31 +17,85 @@
...
+     * We have two options here. Either double the sampling period, or apply a filter to record only the
+     * Nth sample passed to sample_set. By using a filter and keeping the sampling period the same, we avoid
+     * spilling over into the next task.
+     */
+    public function apply_doubling() {
+        $this->filterrate *= 2;
+        $this->samples = array_values(
+            array_filter(
+                $this->samples,
+                function($key) {
```
brendanheywood, line 94: I'm not sure this algorithm is correct. We are removing every second entry in the list of samples but this isn't exactly the same as halving the samples. Consider this test case with 112 time slices:

```
1> a > b > c (1 sample)
2> a > b > c (1 sample)
3> a > b > c (10 samples)
4> a > b > c (100 samples)
```
 if we remove every second record then we could end up with this only 11 time slices 

```
1> a > b > c (1 sample)
3> a > b > c (10 samples)
```

but this isn't what we want, what we really want is this with 56 time slices = 112 * .5:

```
1> a > b > c (1 sample)
3> a > b > c (50 samples)
4> a > b > c (5 samples)
```

keevan: Also I wonder if `apply_doubling` should be a private method. Aside from it being used in unit tests, do you think this would ever be used outside of the logic when adding a new sample?

keevan: @brendanheywood From my read it seems like the samples would be added one at a time, so the removal of every 2nd happens as part of the samples being added instead of them being added as a batch, then processed afterwards. But @jaypha feel free to correct me if this is wrong.

tests/tool_excimer_sample_set_test.php
keevan: Ah sorry add_sample function, I've moved the comment.

PR #171, opened 2022-02-28
Issue # 170: Fix string def for field_userid

jaypha:


lang/en/tool_excimer.php
```diff
@@ -83,7 +83,7 @@
 $string['field_group'] = 'Group';
 $string['field_created'] = 'Created';
 $string['field_finished'] = 'Finished';
-$string['field_user'] = 'User';
+$string['field_userid'] = 'User';
```
keevan, line 86: But field_user is currently being used in other places too (e.g. mustache template)? I think this should be an additional string, not a replacement.

See below:
```php
# admin/tool/excimer/templates/flamegraph.mustache
                    <th>{{#str}} field_user, tool_excimer {{/str}}</th>
                    <th>{{#str}} field_useragent, tool_excimer {{/str}}</th>
# admin/tool/excimer/classes/privacy/provider.php
                'userid' => 'field_userid',
                'useragent' => 'field_useragent',
# admin/tool/excimer/lang/en/tool_excimer.php
$string['field_user'] = 'User';
$string['field_useragent'] = 'User agent';
```

keevan (changes requested):

lang/en/tool_excimer.php
jaypha: The field used to be called user, but changed to userid. The mustache template refers to the same thing, so it was changed.
A quick global search revealed no more instances of 'field_user'.

keevan: Yeah, the problem is not all the fields are set as the full string name, and are used in dynamic strings which makes these sorts of things harder to detect from a grep search.

```
Invalid get_string() identifier: 'field_user' or component 'tool_excimer'. Perhaps you are missing $string['field_user'] = ''; in /var/www/39/admin/tool/excimer/lang/en/tool_excimer.php?

    line 353 of /lib/classes/string_manager_standard.php: call to debugging()
    line 7072 of /lib/moodlelib.php: call to core_string_manager_standard->get_string()
    line 73 of /admin/tool/excimer/classes/profile_table.php: call to get_string()
    line 47 of /admin/tool/excimer/classes/profile_table_page.php: call to tool_excimer\profile_table->make_columns()
    line 51 of /admin/tool/excimer/slowest.php: call to tool_excimer\profile_table_page::display()
```
The offending section is at the table class, e.g. (`admin/tool/excimer/classes/profile_table.php`)

keevan (changes requested):

jaypha:
Alright. I've done the unit test. I've done a search on the letter sequence 'user', and checked each line. I've examined every page in the plugin. If you can still find an issue with 'field_user', I'll give you a prize.

keevan:
> Alright. I've done the unit test. I've done a search on the letter sequence 'user', and checked each line. I've examined every page in the plugin. If you can still find an issue with 'field_user', I'll give you a prize.

Hi @jaypha , when you've updated the column `user` to `userid`, this caused the handling of the user to no longer be handled by the function here https://github.com/catalyst/moodle-tool_excimer/blob/ff09eaa/classes/profile_table.php#L251 and causes this output below:

URL: `/admin/tool/excimer/slowest.php`

Actual:

image

Expected: (user name and a link to the user)

image

Do I still get a prize? :smiling_face_with_tear:

Issue #173, opened 2022-03-14
Be able to track memory usage across a profile and make memory usage a new reason


brendanheywood:
From a high level business point of view we want to see which pages are consistently using lots of memory, specific profiles of ones which use lots, and when looking at the profile a graph showing the memory usage so we can clearly see when the memory spiked up

* [x] for each sample in a profile, also throw in another data point which is the memory used
* [x] add this as a second graph under the profile and lined up so we can see at what point memory spikes up

PR #193, opened 2022-04-11
Add memory usage to the saved profiles

keevan:
- refactor: adjust sample_set class to not require 3rd param by default
- feat: add memusage graph and split sections making it more maintainable
- feat: add memusage storage during batch processing
- fix: line chart display
- fix: amount of memory usage samples shown on the x-axis
- style: cleanup usage in web processor and for json endpoint
- style: further cleanup of cron and web processor files
- feat: add memoryusage sampleset handling to cron processor
- feat: add max memory usage field to be stored on profile save

- [x] Clean up template file descriptions
- [x] Rebase and correct versioning, probably compact the changes into one savepoint.

Example: a long running sleeping web request (e.g. with sleeps at the end)

image

Related to #173 (not does not completely close it)

### DIFF
--- a/templates/memoryusagegraph.mustache
+++ b/templates/memoryusagegraph.mustache
@@ -89,7 +89,35 @@ div.tooltip {
 </style>
 
 <script type="text/javascript">
+
+/**
+ * Generate the display for file size
+ * @param int size The size to convert to human readable form
+ * @return string
+ */
+const display_size = function(size) {
+    // This is snippet of code (with some changes) is from the display_size function in moodlelib.
+    var gb = "{{#str}} sizegb, moodle {{/str}}",
+        mb = "{{#str}} sizemb, moodle {{/str}}",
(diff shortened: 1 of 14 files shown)


templates/memoryusagegraph.mustache
```diff
@@ -27,33 +27,10 @@
...
-    * cookies - Is cookies enabled?
-    * buffering - Is buffering enabled?
-    * fullname - Full nane of user represented by userid.
-    * userlink - Link to user's profile page, or null if no user.
-    * tabs - Tabs markup.
 
     Example context (json):
     {
         "id" : 3,
-        "request" : "/my/index.php/another/deeper/path?time=day&duration=week",
```
brendanheywood: why are these chunks deleted?

keevan: Because the template in question does not use the context object described. For instance, this is the mem usage graph, but the request field is only used on the profile template.

Issue #174, opened 2022-03-21
Test to make sure this picks up runaway recursion issues


brendanheywood:
We had an issue which we think we have solved here:

https://tracker.moodle.org/browse/MDL-59681

I would like to make a minimal repro test case and see if and how excimer would have picked this up. This issue is masked in dev because xdebug detects the recursion and throws an exception but in most prod setups this doesnt happen. But we could actually get excimer to detect deep recursion and log it, and possible go the extra step and stop the script at some level like xdebug does

PR #205, opened 2022-04-13
WIP: Issue #174: Detect and action recursive tasks #199

macgirvin:
WIP: Issue https://github.com/catalyst/moodle-tool_excimer/issues/174 ; Detect recursive tasks

- DB field added 'maxstackdepth' with DB migration tasks [done]
- Calculate the deepest stack depth for each sample set [done]
- flag the resulting profile as "interesting" if the stack depth exceeds a defined minimum. [done]
- optionally(?) kill the offending task [in progress]

### DIFF
--- a/classes/cron_processor.php
+++ b/classes/cron_processor.php
@@ -16,6 +16,8 @@
 
 namespace tool_excimer;
 
+use tool_excimer\script_metadata;
+
 /**
  * Class for processing cron and adhoc tasks.
  *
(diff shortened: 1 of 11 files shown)


classes/sample_set.php
```diff
@@ -66,6 +78,16 @@ public function add_sample(\ExcimerLogEntry $sample) {
...
+        // Each time a sample is added, recalculate the maxstackdepth for this set.
+        foreach ($this->samples as $sample) {
+            $trace = $sample->getTrace();
+            if ($trace) {
+                $stackdepth = count($trace);
+                if ($stackdepth > $this->maxstackdepth) {
+                    $this->maxstackdepth = $stackdepth;
+                }
+            }
+        }
```
keevan, line 90: I believe you don't need to look through all the samples again to find the maximum. You'll only need to look at the current maximum and the current sample (to be added).

I also think instead of doing 
```
                if ($stackdepth > $this->maxstackdepth) {
```
You can probably get by with something like `$this->maxstackdepth = max($this->maxstackdepth, count($trace)`; (keeps nesting to a minimum)

classes/cron_processor.php
```diff
@@ -142,12 +144,18 @@ public function findtaskname(\ExcimerLogEntry $sample): ?string {
      * @throws \dml_exception
      */
     public function process(manager $manager, float $finishtime): void {
+        $reasonstack = 0;
         $duration = $finishtime - $this->currenttask->starttime;
+
+        if (property_exists($this, 'currenttask') && $this->currenttask
```
brendanheywood, line 150: can you move this logic into get_reasons() it should be fairly similar to 

$this->is_considered_slow($profile)

ala:

$this->is_considered_deep($profile)

PR #208, opened 2022-04-13
 WIP: Issue #174: Detect and action recursive tasks

macgirvin:
WIP: Issue #174: Detect and action recursive tasks

branch GH174a
-    DB field added 'maxstackdepth' with DB migration tasks [done]
-    Calculate the deepest stack depth for each sample set [done]
-    flag the resulting profile as "interesting" if the stack depth exceeds a defined minimum. [done]
-    optionally(?) kill the offending task [in progress]


classes/cron_processor.php
```diff
@@ -142,12 +144,18 @@ public function findtaskname(\ExcimerLogEntry $sample): ?string {
      * @throws \dml_exception
      */
     public function process(manager $manager, float $finishtime): void {
+        $reasonstack = 0;
         $duration = $finishtime - $this->currenttask->starttime;
+
+        if (property_exists($this, 'currenttask') && $this->currenttask
```
brendanheywood: can you please move this logic to inside get_reasons()

classes/web_processor.php
```diff
@@ -32,7 +32,7 @@ class web_processor implements processor {
     protected $profile;
 
     /** @var sample_set */
-    protected $sampleset;
+    public $sampleset;
```
keevan: Was there any particular reason this is now a **public** scoped variable instead of staying protected?

classes/sample_set.php
```diff
@@ -66,6 +78,16 @@ public function add_sample(\ExcimerLogEntry $sample) {
             $this->samples[] = $sample;
             $this->counter = 0;
         }
+        // Each time a sample is added, recalculate the maxstackdepth for this set.
+        foreach ($this->samples as $sample) {
+            $trace = $sample->getTrace();
+            if ($trace) {
+                $stackdepth = count($trace);
+                if ($stackdepth > $this->maxstackdepth) {
```
brendanheywood: there was a comment here which was lost from kevin around using max($a, $b) to make this a little simpler

classes/sample_set.php
```diff
@@ -66,6 +78,16 @@ public function add_sample(\ExcimerLogEntry $sample) {
...
+                    $this->maxstackdepth = $stackdepth;
+                }
+            }
+        }
```
keevan: I believe you don't need to look through all the samples again to find the maximum. You'll only need to look at the current maximum and the current sample (to be added).

I also think instead of doing
```
                if ($stackdepth > $this->maxstackdepth) {
```
You can probably get by with something like `$this->maxstackdepth = max($this->maxstackdepth, count($trace)`; (keeps nesting to a minimum)

classes/sample_set.php
```diff
@@ -52,6 +53,17 @@ public function __construct(string $name, float $starttime, int $samplelimit) {
         $this->samplelimit = $samplelimit;
     }
 
+    /**
+     * Return the stack depth for this set.
+     *
+     * @param \ExcimerLogEntry $sample
+     * @return void
+     */
+
```
keevan: Please remove this unnecessary line

classes/web_processor.php
brendanheywood: does this need to be public?

classes/sample_set.php
```diff
@@ -52,6 +53,17 @@ public function __construct(string $name, float $starttime, int $samplelimit) {
         $this->samplelimit = $samplelimit;
     }
 
+    /**
+     * Return the stack depth for this set.
+     *
+     * @param \ExcimerLogEntry $sample
```
keevan: No params need to be provided so it does not need the @param hint here.

classes/web_processor.php
```diff
@@ -80,12 +80,16 @@ public function get_min_duration(): float {
...
      */
     public function process(manager $manager, bool $isfinal) {
+        $reasonstack = 0;
         $log = $manager->get_profiler()->flush();
         $this->sampleset->add_many_samples($log);
 
         $current = microtime(true);
         $this->profile->set('duration', $current - $manager->get_starttime());
-        $reason = $manager->get_reasons($this->profile);
+        if ($this->sampleset->get_stack_depth() > script_metadata::get_stack_limit()) {
```
brendanheywood: this should be somehwat redundant when the logic is moved into get_reasons()

keevan (commented):
Had a look, mostly good. It also seems to be missing a setting in the GUI to customise the stack limit.

classes/web_processor.php
macgirvin: I added a function to sample_set to retrieve the stored maximum, and couldn't access it from the web_processor with protected scope on the sample_set. I'll rework this. It was one part of the attempt to provide as much of a common interface as possible between the completely dissimilar processors.

macgirvin:
> Had a look, mostly good. It also seems to be missing a setting in the GUI to customise the stack limit.

:facepalm: Forgot to click 'save' on the settings stuff.

classes/web_processor.php
```diff
@@ -32,6 +32,7 @@ class web_processor implements processor {
     protected $profile;
 
     /** @var sample_set */
+
```
brendanheywood, line 35: extra whitespace

db/install.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<XMLDB PATH="admin/tool/excimer/db" VERSION="20220411" COMMENT="XMLDB file for Moodle admin/tool/excimer"
+
```
brendanheywood, line 2: extra whitespace, did you use the xml db editor to edit this file?

macgirvin:
Closing to re-submit after squashing and fixing conflicts.

Issue #203, opened 2022-04-13
Bug with missing userid's


brendanheywood:
/admin/tool/excimer/slowest.php

Here is a bunch of profiles, all of these are triggered by the link crawler and you can see this in the profile details:

![image](https://user-images.githubusercontent.com/187449/163090196-fea5aa8d-3d39-4426-a391-b597af87230f.png)
![image](https://user-images.githubusercontent.com/187449/163090299-577c477a-b030-4206-8a55-77776a830ec4.png)

I'm pretty sure that excimer is picking up the USER->id too early and not updating it when the user is actually setup later in the script. So I think just check it again a second time at the end of the shutdown handler

PR #210, opened 2022-04-14
Excimer issue #203, missing userid on crawler requests

macgirvin:
Excimer #203; bug with missing user id's

set userid at end of shutdown handler (web only, cron_processor currently has no profile accessible to shutdown processor)

### DIFF
--- a/classes/profile.php
+++ b/classes/profile.php
@@ -193,6 +193,12 @@ class profile extends persistent {
         $this->raw_set('timemodified', $now);
         $this->raw_set('usermodified', $USER->id);
 
+        // We may not have obtained a valid userid when the profile record was created.
+        // If the current profile userid is 0, but there's a valid $USER->id, update it.
+        if ($USER->id && ! $this->raw_get('userid')) {
+            $this->raw_set('userid', $USER->id);
+        }
+
         if ($this->raw_get('id') <= 0) {
             $this->raw_set('timecreated', $now);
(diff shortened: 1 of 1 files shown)


classes/web_processor.php
```diff
@@ -41,6 +41,7 @@ class web_processor implements processor {
      * @param manager $manager The profiler manager object
      */
     public function init(manager $manager) {
+        global $USER;
```
keevan: This does not appear to be used anywhere in the method, please consider removing it.

brendanheywood:
Looks good to me

classes/profile.php
```diff
@@ -191,6 +191,12 @@ public function save_record(): int {
         $this->raw_set('timemodified', $now);
         $this->raw_set('usermodified', $USER->id);
 
+        // We may not have obtained a valid userid when the profile record was created.
+        // If the current profile userid is 0, but there's a valid $USER->id, update it.
+        if ((! (int) $this->raw_get('userid')) && (int) $USER->id) {
+            $this->raw_set('userid', (int) $USER->id);
+        }
```
keevan: There's a lot of casting here :magic_wand: 

Are these necessary? And if not, can we please remove them as they add extra cognitive overhead. I think the user->id check does not need it. Also in the raw_get check for userid, it always returns a value or null if it's not set. I believe an isset check here will be sufficient.

PR #215, opened 2022-04-19
Issue #174: Detect and action recursive tasks

macgirvin:
Issue #174: Detect and action recursive tasks
  - DB field added 'maxstackdepth' with DB migration tasks [done]
  - Calculate the deepest stack depth for each sample set [done]
  - flag the resulting profile as "interesting" if the stack depth exceeds a defined minimum. [done]

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -324,5 +324,18 @@ function xmldb_tool_excimer_upgrade($oldversion) {
         // Excimer savepoint reached.
         upgrade_plugin_savepoint(true, 2022041301, 'tool', 'excimer');
     }
+
+    if ($oldversion < 2022042000) {
+        // Add field maxstackdepth.
+        $table = new xmldb_table('tool_excimer_profiles');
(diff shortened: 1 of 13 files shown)


db/upgrade.php
```diff
@@ -324,5 +324,18 @@ function xmldb_tool_excimer_upgrade($oldversion) {
         // Excimer savepoint reached.
         upgrade_plugin_savepoint(true, 2022041301, 'tool', 'excimer');
     }
+
+    if ($oldversion < 2022041400) {
+        // Add field maxstackdepth.
+        $table = new xmldb_table('tool_excimer_profiles');
+        $field = new xmldb_field('maxstackdepth', XMLDB_TYPE_INTEGER, '11', null, XMLDB_NOTNULL, null, 0, 'userid');
+
+        if ($dbman->table_exists($table) && $dbman->field_exists($table, $field)) {
```
keevan: I think you meant `!$dbman->field_exists($table, $field)`, since this would otherwise skip it from being added if it does not yet exist.

keevan: Also the table_exists check is somewhat superfluous, as the previous upgrade blocks do not include this and even if it does not exist, you haven't tried to correct it by adding the table either.

Issue #138, opened 2022-01-05
Refine logic for moodle user for task profiles


brendanheywood:
A cron task can set the user and keep changing it, eg when sending a forum email it sudo's as the person sending it. Either we can correctly determine the current user(s) which might be as simple as detecting if the user was swapped to something which isn't the admin user at any point during the task, OR we just clearly say that we can't detect it correctly and don't show misleading info

Logic:
1) if we have a user, and there currently isn't one, then set the user
2) if we have a user who isn't the guest, and the current one is guest, then set the user
3) if we have a user who isn't the cli admin user, and the current is the cli admin user then override it

PR #216, opened 2022-04-19
Issue #138; refine logic for moodle user for task profiles

macgirvin:
Issue #138; This is a slight rework of the previous patch for setting userid (PR #210), which should resolve both this issue and issue #203 .

### DIFF
--- a/classes/profile.php
+++ b/classes/profile.php
@@ -193,9 +193,7 @@ class profile extends persistent {
         $this->raw_set('timemodified', $now);
         $this->raw_set('usermodified', $USER->id);
 
-        // We may not have obtained a valid userid when the profile record was created.
-        // If the current profile userid is 0, but there's a valid $USER->id, update it.
-        if ($USER->id && ! $this->raw_get('userid')) {
+        if ($this->check_update_userid($USER->id)) {
             $this->raw_set('userid', $USER->id);
         }
 
@@ -227,6 +225,37 @@ class profile extends persistent {
         }
 
         return (int) $this->raw_get('id');
+    }
+
+    /**
+     * Decide if the user id stored with this profile should be updated with the current $USER->id.
+     *
+     * @param int $currentid
+     * @return bool
+     */
+    protected function check_update_userid(int $currentid) : bool {
+        global $CFG;
+
+        $stored = (int) $this->raw_get('userid');
+
+        // We may not have obtained a valid userid when the profile record was created.
+        // If the stored userid is 0, and there's now a valid $USER->id, update the stored userid.
+        if ($currentid && !$stored) {
+            return true;
+        }
+
+        // If current userid isn't guest, but stored user is guest, update it.
+        $guestid = (int) $CFG->siteguest;
+        if ($guestid && $currentid !== $guestid && $stored === $guestid) {
+            return true;
+        }
+
+        // If current user isn't cli admin, but stored user is cli admin, update it.
+        $adminid = get_admin()->id;
+        if ($adminid && $currentid !== $adminid && $stored === $adminid) {
+            return true;
+        }
+        return false;
     }
 
     /**


classes/profile.php
```diff
@@ -194,8 +194,9 @@ public function save_record(): int {
         $this->raw_set('usermodified', $USER->id);
 
         // We may not have obtained a valid userid when the profile record was created.
-        // If the current profile userid is 0, but there's a valid $USER->id, update it.
-        if ($USER->id && ! $this->raw_get('userid')) {
+        // If the current $USER->id is non-zero, update it unless the profile userid is admin.
+        $profileuser = (int) $this->raw_get('userid');
+        if ($USER->id && $profileuser !== 1) {
```
brendanheywood: What is the back story here? 1 looks like a magic number to me, perhaps instead you should be leveraging is_guest() ?

macgirvin: I should have asked but I thought I remembered that admin is always userid 1 in Moodle. I'll look for a call to find the admin uid.

macgirvin: Google says 2, so I'll look into this deeper and correct it.

macgirvin: Looks like guest is 1 and admin is 2. Now, I just need to find out if that's always the case, or how to find the admin userid if it's not.

brendanheywood: The admin id can be anything and can change over time. But I would like to understand the intent of the code first and that should be reflected in the comment, and then worry about the details

classes/profile.php
```diff
@@ -193,12 +193,32 @@ public function save_record(): int {
...
+        // If current userid isn't guest, but stored user is guest, update it.
+        $guestrole = get_guest_role();
+        $guestid = (int) $guestrole->id;
+        if ($USER->id !== $guestid && $storeduser === $guestid) {
             $this->raw_set('userid', $USER->id);
         }
 
+        // If current user isn't cli admin, but stored user is cli admin, update it.
+        if (script_metadata::get_script_type() === self::SCRIPTTYPE_CLI) {
+            $admins = get_admins();
```
keevan: I think you can use `get_admin` here instead, and it will be the main cli user, no loop required.

classes/profile.php
```diff
@@ -193,12 +193,32 @@ public function save_record(): int {
...
+
+        // If current userid isn't guest, but stored user is guest, update it.
+        $guestrole = get_guest_role();
+        $guestid = (int) $guestrole->id;
+        if ($USER->id !== $guestid && $storeduser === $guestid) {
             $this->raw_set('userid', $USER->id);
         }
 
+        // If current user isn't cli admin, but stored user is cli admin, update it.
+        if (script_metadata::get_script_type() === self::SCRIPTTYPE_CLI) {
```
keevan: I don't think we should scope this only for CLI scripts. It probably happens more often in CLI scripts but could happen via web based ones too?

classes/profile.php
```diff
@@ -193,12 +193,32 @@ public function save_record(): int {
...
-        // If the current profile userid is 0, but there's a valid $USER->id, update it.
-        if ($USER->id && ! $this->raw_get('userid')) {
+        // If the stored userid is 0, and there's now a valid $USER->id, update the stored userid.
+        if ($USER->id && !$storeduser) {
+            $this->raw_set('userid', $USER->id);
+        }
+
+        // If current userid isn't guest, but stored user is guest, update it.
+        $guestrole = get_guest_role();
+        $guestid = (int) $guestrole->id;
```
brendanheywood: I'm not sure this is right, this is the id of the role, not the id of the guest user. Instead I would not even use the guest id I'd just call is_guest() or isguestuser()

classes/profile.php
```diff
@@ -193,12 +193,32 @@ public function save_record(): int {
...
+        if ($USER->id !== $guestid && $storeduser === $guestid) {
             $this->raw_set('userid', $USER->id);
         }
```
keevan: I feel like for these 3 conditional blocks, the outcome is the same (if any were true), i.e. it will update the userid on the profile.

I see potential overlap in checks, e.g. if you pass the first check, you're always checking for guest role, and if you were in CLI, you're always checking admin users. 

It might be better as one big if condition or calculated in a separate method/function and set if required. Personally I'm leaning towards handling this in a separate private method, maybe something like `get_resolved_user` which returns the userid and updates it if its different.

classes/profile.php
brendanheywood: get_admin() returns the 'primary admin' or main admin. They are the only one which is treated special, the other admins are not a concern. So we don't need the loop over get_admins()

I know this is confusing moodle internals :face_with_spiral_eyes:

classes/profile.php
```diff
@@ -193,12 +193,32 @@ public function save_record(): int {
...
         // We may not have obtained a valid userid when the profile record was created.
-        // If the current profile userid is 0, but there's a valid $USER->id, update it.
-        if ($USER->id && ! $this->raw_get('userid')) {
+        // If the stored userid is 0, and there's now a valid $USER->id, update the stored userid.
+        if ($USER->id && !$storeduser) {
+            $this->raw_set('userid', $USER->id);
+        }
+
+        // If current userid isn't guest, but stored user is guest, update it.
+        $guestrole = get_guest_role();
```
keevan: I also don't think this is the correct check. This returns the roleid for guest users, and I think it the value to check here might need to be `$CFG->siteguest` or similar instead

PR #226, opened 2022-04-27
Issue #227: excimer field modification to ensure DB consistency.

macgirvin:
Github #227 excimer field modification to ensure DB consistency.
Field 'maxstackdepth' created with no default but added with a default.
This checkin ensures that if the field exists already with no default
that it will be modifed to set default to 0.

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -325,17 +325,19 @@ function xmldb_tool_excimer_upgrade($oldversion) {
         upgrade_plugin_savepoint(true, 2022041301, 'tool', 'excimer');
     }
 
-    if ($oldversion < 2022042000) {
+    if ($oldversion < 2022042700) {
         // Add field maxstackdepth.
         $table = new xmldb_table('tool_excimer_profiles');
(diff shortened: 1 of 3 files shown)


db/upgrade.php
```diff
@@ -325,17 +325,19 @@ function xmldb_tool_excimer_upgrade($oldversion) {
         upgrade_plugin_savepoint(true, 2022041301, 'tool', 'excimer');
     }
 
-    if ($oldversion < 2022042000) {
+    if ($oldversion < 2022042700) {
```
brendanheywood, line 328: The 2022042000 version is baked in and already out in the wild (even if only on a very small number of sites). Best practice is to never ever modify a previous versions upgrade step as you can't know which sites have or havent run that step. Instead you need to make future steps work with whatever state may be present.

So leave the existing one alone and make a new if ($oldversion < 2022042700) { section

brendanheywood:
Note this is sorted in https://github.com/catalyst/moodle-tool_excimer/pull/228

Issue #227, opened 2022-04-27
maxstackdepth field inconsistency


macgirvin:
The maxstackdepth field definition isn't consistent between install.xml and upgrade.php. This can lead to database schema alignment issues.

PR #228, opened 2022-04-27
Issue #227; excimer field modification to ensure DB consistency.

macgirvin:
Issue #227; excimer field modification to ensure DB consistency.
Field 'maxstackdepth' created with no default but added with a default.
This checkin ensures that if the field exists already with no default
that it will be modifed to set default to 0.


brendanheywood:
Please push this into https://github.com/catalyst/moodle-tool_excimer/pull/226

PR #241, opened 2022-05-05
fix: install.xml for MOODLE_35_STABLE support

### LINKED ISSUE #236: Fix the schema check
tool_excimer_profiles
column 'flamedatad3' is not expected (B)
column 'memoryusagedatad3' is not expected (B)

keevan:
Fixed #236


jaypha:
Does this need additions to upgrade.php for existing databases?

keevan:
> Does this need additions to upgrade.php for existing databases?

It's a bit hard to say.

https://github.com/moodle/moodle/blob/1a744030d639f93f2a147d40f0a38572fa0f846d/lib/xmldb/xmldb_field.php#L329 suggests that the length is simply removed. 

That said, I'll add it to the upgrade script for consistency, but I'm not certain if it will change the types internally. 

A summary of the types is here: https://docs.moodle.org/dev/XMLDB_column_types with MySQL listing 3 possible types it could ultimately apply.

jaypha:
Alright. We'll see if that works.

PR #256, opened 2022-09-02
Issue #200: Add duration to tooltip

### LINKED ISSUE #200: When you hover over a section of the flame graph show the duration of the segment
Sometimes it can be hard to quickly grok the scale of how long something took, as it is a fraction of a duration.

When you hover over a flame I'd like to see the in hover text under the flame something like '100ms' or 1.2 secs'. This would just be an estimate, basically the number of samples X the sample rate so there will be some error in it and thats ok. 

https://github.com/spiermar/d3-flame-graph#setDetailsHandler
https://github.com/spiermar/d3-flame-graph#label
https://github.com/spiermar/d3-flame-graph#tooltip

petersistrom:
Resolves #200


jaypha:
I'm curious as to why you replaced d3-flamegraph.min.js with the other two.

petersistrom:
Good question. Was an artifact of debugging I think. Updated

PR #257, opened 2022-09-02
Issue #249: locking profiles

### LINKED ISSUE #249: Mark a profile as locked so it will not be deleted
Make this a text area with a reason to lock it. If empty then unlocked.

We use this to put a link to a wr / tracker / ticket etc

jaypha:
Resolves #249

### DIFF
--- a/lock_profile.php
+++ b/lock_profile.php
@@ -52,6 +52,7 @@ $form = new \tool_excimer\form\lock_reason_form($url);
 
 if ($data = $form->get_data()) {
     $DB->update_record('tool_excimer_profiles', (object)['id' => $profileid, 'lockreason' => trim($data->lockreason)]);
+    redirect(new moodle_url('/admin/tool/excimer/profile.php', ['id' => $profileid]));
 } else {
     $form->set_data(['lockreason' => $DB->get_field('tool_excimer_profiles', 'lockreason', ['id' => $profileid])]);
 }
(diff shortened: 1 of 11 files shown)


lock_profile.php
```diff
@@ -0,0 +1,89 @@
...
+$tabs = new tabs($url);
+echo $output->render_tabs($tabs);
+
+$responsecode = helper::status_display($profile->get('scripttype'), $profile->get('responsecode'));
+$method = $profile->get('method');
+$request = $profile->get('request');
+echo html_writer::tag('h3', "$responsecode $method $request");
+
+if ($data) {
+    echo $output->notification(get_string('profile_updated', 'tool_excimer'), 'success');
```
brendanheywood: After you lock it it should go back to the page you were on

templates/profile.mustache
```diff
@@ -161,6 +161,10 @@
                     <th>{{#str}} field_dbreplicareads, tool_excimer {{/str}}</th>
                     <td>{{dbreplicareads}}</td>
                 </tr>
+                <tr>
+                    <th>Lock reason</th>
```
brendanheywood: 1) this should nbe a lang string

2) can you move this out of the table and put it under just as text. Can you run it through the output api to turn any urls into links:

image

profile.php
```diff
@@ -102,6 +109,9 @@
 $deleteallbutton = new \single_button($deleteallurl, get_string('deleteprofiles_script', 'tool_excimer'));
 $deleteallbutton->add_confirm_action(get_string('deleteprofiles_script_warning', 'tool_excimer'));
 
+$lockprofileurl = new \moodle_url('/admin/tool/excimer/lock_profile.php', ['profileid' => $profileid]);
+$lockprofilebutton = new \single_button($lockprofileurl, 'Lock/Unlock', 'GET');
```
brendanheywood, line 114: 1) should be lang string
2) Can you make this say 'Lock' when it is not locked and Unlock with it is locked

templates/profile.mustache
brendanheywood: Actually can you put this at the top, and put it inside a notification box saying 'This is locked' with the reason text under that, and then the table

PR #262, opened 2022-09-06
Issue #244: Add before_session_start to include waiting for session locks

### LINKED ISSUE #244: Excimer isn't correctly including the time waiting for a session lock
This is hard, because we are relying on the _after_config hook and this delay is in the session manager before this hook is called.

It gives misleading numbers on pages where there could be say a 10 second delay waiting and then 1 second of processing, and excimer shows it as just one second. 

I think at minimum lets try and infer the time the script started and just show this in the profile metadata even if it isn't in the flame graph.

The tideways profiler is hard coded to allow it to kick off much earlier:
https://github.com/moodle/moodle/blob/master/lib/setup.php#L442

A proper solution might be:
* [x] land this tracker: https://tracker.moodle.org/browse/MDL-75014
* [ ] document backporting that + patch in ci for this plugin
* [ ] start excimer in that callback

Good test is using this:
https://master.localhost/admin/tool/sessionbreaker/

petersistrom:
Resolves #244

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,14 @@ git clone git@github.com:catalyst/moodle-tool_excimer.git admin/tool/excimer
 ```
 
 Then login as admin (it should detect the extension), and click through the upgrade process.
+## Applying core patches
 
+For Moodle versions lower than 4.1 this plugin requires Moodle tracker MDL-75014 to be backported to maintain the plugin functionality.
+#### Moodle 3.5 - 4.0:
+Apply the patch:
+<pre>
+git am --whitespace=nowarn < admin/tool/excimer/patch/MOODLE_35_STABLE.diff
+</pre>
 ## Troubleshooting
 
 **ExcimerProfiler class does not exist**.
--- a/lib.php
+++ b/lib.php
@@ -27,6 +27,15 @@ use tool_excimer\manager;
 use tool_excimer\check\slowest;
 
 /**
+ * Hook to run plugin before session start.
+ */
+function tool_excimer_before_session_start() {
+    // Start plugin.
+    $manager = manager::get_instance();
+    $manager->init();
+}
+
+/**
  * Hook to be run after initial site config.
  *
  * This allows the plugin to selectively activate the ExcimerProfiler while
@@ -37,11 +46,9 @@ use tool_excimer\check\slowest;
  * See also https://docs.moodle.org/dev/Login_callbacks#after_config.
  */
 function tool_excimer_after_config(): void {
-    // Skip setup if unit testing, but do the test here to avoid complicating manager::is_profiling().
-    if (!PHPUNIT_TEST && manager::is_profiling()) {
-        $manager = manager::create();
-        $manager->init();
-    }
+    // Start processor.
+    $manager = manager::get_instance();
+    $manager->start_processor();
 }
 
 /**
--- a/classes/manager.php
+++ b/classes/manager.php
@@ -47,6 +47,8 @@ class manager {
     private $timer;
     /** @var float */
     private $starttime;
+    /** @var manager */
+    private static $instance;
 
     /**
      * Generates the samples for the script.
@@ -76,6 +78,19 @@ class manager {
     }
 
     /**
+     * Get manager instance.
+     *
+     * @return manager
+     * @throws \dml_exception
+     */
+    public static function get_instance(): manager {
+        if (!self::$instance) {
+            self::create();
+        }
+        return self::$instance;
+    }
+
+    /**
      * Constructs the manager.
      *
      * @param processor $processor The object that processes the samples generated.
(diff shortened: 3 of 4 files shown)


lib.php
```diff
@@ -52,3 +48,11 @@ function tool_excimer_after_config(): void {
 function tool_excimer_performance_checks(): array {
     return [new slowest()];
 }
+
+/**
+ * Hook to run plugin before session start.
+ */
+function tool_excimer_before_session_start() {
+    // Start plugin.
+    tool_excimer_after_config();
```
brendanheywood: because you have refactored it I would call this the same here manager::create_if_needed();

brendanheywood: I'd also move them side by side, or better yet in the order they will be executed so this one first

classes/web_processor.php
```diff
@@ -98,7 +99,7 @@ public function process(manager $manager, bool $isfinal) {
         $this->profile->set('duration', $current - $manager->get_starttime());
         $this->profile->set('maxstackdepth', $this->sampleset->get_stack_depth());
         $reason = $manager->get_reasons($this->profile);
-        if ($reason !== profile::REASON_NONE) {
+        if ($reason !== profile::REASON_NONE && $USER != null) {
```
brendanheywood: This seems wrong? Why would we not want to save the record if we don't have a user yet? Is this related to this patch?

petersistrom: To test this I was putting a sleep() on one page to lock the session and then trying to load another at the same time. I was getting this error when the second page loaded, as it was trying to save a null user. 

image

lib.php
petersistrom: Updated

classes/web_processor.php
jaypha: This could be because it is trying to save user info before the session has started, therefore no user information is available.

jaypha:
Creating the manager as is before the session starts will cause problems as not all information is available yet (e.g. $USER). The manager needs to start the profiler and timer while in `tool_excimer_before_session_start` but not flesh out the details until `tool_excimer_after_config`.

classes/web_processor.php
brendanheywood: As an explicit test scenario I'd like you to test a page which is holding the session lock open in one page for say 10 seconds, and then we are profiling a second page which is blocked by the first page. If you set the numbers really low so that it is saving partial profiles every 1 seconds, then I would expect the profile for page B to exist without any user data after 1 second and to show all of its time spent in the session waiting for a lock. Then after 10 seconds the lock will be released, the user will be logged in, and the profile metadata will be updated to include the user that we now know who they are

jaypha:
I'm fine with this. Not sure if it should be merged now, or wait until the tracker lands and is backported.

brendanheywood:
Lets wait just a little, I'm hopeful the tracker will land soon. It isn't in this weeks batch but should be in the next week

lib.php
```diff
@@ -37,11 +46,9 @@
...
  */
 function tool_excimer_after_config(): void {
-    // Skip setup if unit testing, but do the test here to avoid complicating manager::is_profiling().
-    if (!PHPUNIT_TEST && manager::is_profiling()) {
-        $manager = manager::create();
-        $manager->init();
-    }
+    // Start processor.
+    $manager = manager::get_instance();
+    $manager->start_processor();
```
brendanheywood, line 51: Why is the logic for what _session_start is doing different to what _after config is doing? I would assume they should be identical

jaypha:
The trackers have been backported.

keevan (changes requested):
As mentioned recently the tracker https://tracker.moodle.org/browse/MDL-75014 has landed and so this PR will also need to include the patch for the tracker.

Once that's done and remaining review items addressed it can be merged.

README.md
```diff
@@ -124,7 +124,32 @@ git clone git@github.com:catalyst/moodle-tool_excimer.git admin/tool/excimer
...
+git checkout -b newbranch
+
+// Cherry-pick MDL-75014
+git cherry-pick 519c932cd63749843dbbb2af883793d5c937ee0c
+
+// Resolve any merge conflicts only accepting changes for the tracker (changes to upgrade.txt were omitted for simplicity when backporting) 
+git cherry-pick --continue
+
+// Create the patch
+git format-patch MOODLE_310_STABLE --stdout > admin/tool/excimer/patch/MOODLE_35_STABLE.diff
```
keevan: This is good but I don't believe this is required here. It might be useful in the wiki / reusable workflows repository if not already.

lib.php
petersistrom: This logic had to be separated out to start the timer `tool_excimer_before_session_start` and then flesh the details in `tool_excimer_after_config` as explained here https://github.com/catalyst/moodle-tool_excimer/pull/262#issuecomment-1237780878

keevan (approved):

PR #264, opened 2022-09-07
Issue #250: Implement approximate counting algorithm

### LINKED ISSUE #250: Breakdown of all requests and histograms of duration with totals
We want to be able to see a list of all the worst pages, independent of whether they have profiles or not. In order to do this efficiently we want to have probabilistic counting for all numbers involved so that the io is log(n) as you scale.

https://en.wikipedia.org/wiki/Approximate_counting_algorithm

The end result is a table where we can see:
* [ ] a list of all of the urls by 'group' - the same grouping where we strip off params
* [x] these will need to become first class records one per group
* [x] we will have a fuzzy count of how many hits this url group has had
* [x] we will have a fuzzy breakdown count of how many took 1 sec, 2 second, 4 second etc
* [ ] we now we can calculate percentiles and averages and histograms
* [x] we can integrate them to come up with 'seconds spent serving page X' and then store this number itself too
* [x] make all of these numbers normalized to a month, just reset them each month
* [ ] or better yet can we hold this month and last month? the pro rata days in this month against last month

jaypha:
Resolves #250

### DIFF
--- a/classes/manager.php
+++ b/classes/manager.php
@@ -34,6 +34,8 @@ class manager {
     const FLAME_OFF_PARAM_NAME = 'FLAMEALLSTOP';
     /** Don't collect profile for current scripts. */
     const NO_FLAME_PARAM_NAME = 'DONTFLAMEME';
+    /** The size of a PHP integer in bits (64 on Linux x86_64, 32 on Windows). */
+    const PHP_INT_SIZE_BITS = 8 * PHP_INT_SIZE;
 
     /** @var processor */
     private $processor;
@@ -242,5 +244,30 @@ class manager {
         // By this stage, the duration provided should have exceeded the min
         // requirements for all the different timing types (if they exist).
         return true;
+    }
+
+    /**
+     * Increments a counter using the approximate counting algorithm.
+     * Can handle up to PHP_INT_SIZE_BITS counts (2^PHP_INT_SIZE_BITS events).
+     *
+     * See https://en.wikipedia.org/wiki/Approximate_counting_algorithm
+     *
+     * @param int $current The current count.
+     * @return int The new count.
+     */
+    public static function approximate_increment(int $current): int {
+        // If the number of events is ever expected to be more than 2 billion, a refactor may be needed.
+
+        $bits = random_int(PHP_INT_MIN, PHP_INT_MAX);
+
+        // This gives us a number of bits equal to the current count. The rest are all zeroed.
+        $bits = $bits << (self::PHP_INT_SIZE_BITS - $current);
+
+        // If the bits are all zero (equiv to all coin tosses = tails), then we increment the counter.
+        if ($bits == 0) {
+            return $current + 1;
+        }
+
+        return $current;
     }
 }
--- a/classes/output/tabs.php
+++ b/classes/output/tabs.php
@@ -48,7 +48,6 @@ class tabs implements \templatable {
     public function export_for_template(\renderer_base $output): array {
         $tabs = [
             'tabs' => [
-
                 [
                     'id' => 'slowest_web',
                     'link' => [[ 'link' => new \moodle_url('/admin/tool/excimer/slowest_web.php') ]],
@@ -72,6 +71,12 @@ class tabs implements \templatable {
                     'link' => [[ 'link' => new \moodle_url('/admin/tool/excimer/unfinished.php') ]],
                     'title' => get_string('report_unfinished', 'tool_excimer'),
                     'text' => get_string('unfinished', 'tool_excimer'),
+                ],
+                [
+                    'id' => 'page_groups',
+                    'link' => [[ 'link' => new \moodle_url('/admin/tool/excimer/page_groups.php') ]],
+                    'title' => get_string('report_page_groups', 'tool_excimer'),
+                    'text' => get_string('tab_page_groups', 'tool_excimer'),
                 ],
             ],
         ];
--- a/classes/page_group.php
+++ b/classes/page_group.php
@@ -1,0 +1,118 @@
+<?php
+// This file is part of Moodle - https://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_excimer;
+
+use core\persistent;
+
+/**
+ * Metadata for profile groups.
+ *
+ * @package   tool_excimer
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class page_group extends persistent {
+
+    /** The name of the database table. */
+    public const TABLE = 'tool_excimer_page_groups';
+
+    /**
+     * Returns the fuzzy counts for duration.
+     *
+     * @return array
+     */
+    protected function get_fuzzydurationcounts(): array {
+        $raw = $this->raw_get('fuzzydurationcounts');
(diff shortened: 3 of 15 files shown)


classes/output/tabs.php
```diff
@@ -73,6 +72,12 @@ public function export_for_template(\renderer_base $output): array {
                     'title' => get_string('report_unfinished', 'tool_excimer'),
                     'text' => get_string('unfinished', 'tool_excimer'),
                 ],
+                [
+                    'id' => 'page_groups',
+                    'link' => [[ 'link' => new \moodle_url('/admin/tool/excimer/page_groups.php') ]],
```
brendanheywood, line 77: this might end up merged into the existing tabs rather than a new tab, we can refine this toward the end

classes/page_group.php
```diff
@@ -0,0 +1,114 @@
...
+        $median = ($high + $low) / 2.0;
+        $fuzzydurationsum += $median;
+
+        $fuzzydurationcounts[$high] = manager::approximate_increment($fuzzydurationcounts[$high] ?? 0);
+
+        $profilegroup->set('fuzzydurationcounts', $fuzzydurationcounts);
+        $profilegroup->set('fuzzydurationsum', $fuzzydurationsum);
+
+        if ($existing != $profilegroup->to_record()) {
+            $profilegroup->save();
```
brendanheywood: I don't like display logic mutating the data, eventually we want to push this off to the data ware house so the data needs to be ready to go without a human having viewed this report. so either calculate it when it changes which should be not too bad, or worst case in a task but I would like to avoid that

classes/task/purge_profile_groups.php
```diff
@@ -0,0 +1,47 @@
...
+        return get_string('task_purge_profile_groups', 'tool_excimer');
+    }
+
+    /**
+     * Do the job.
+     */
+    public function execute() {
+        global $DB;
+
+        $month = userdate(date('2 months ago'), '%Y%m');
```
brendanheywood: Seeing as you've implemented this as a full history I'm inclined to have the option to keep the data, so can you make this an admin option and make it purge data older than 13 months by default which means at any give time with the partial month in progress we have a full years worth of data. Maybe make the setting say 12 months but add the +1 in code

classes/web_processor.php
```diff
@@ -66,6 +66,7 @@ function () use ($manager) {
                 $manager->get_timer()->stop();
                 $manager->get_profiler()->stop();
                 $this->process($manager, true);
+                page_group::record_fuzzy_counts($this->profile);
```
brendanheywood, line 71: this is called in process above, do we need this here?

classes/page_group.php
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * Records fuzzy count metadata about a page group.
+     *
+     * @param profile $profile The profile to pull the information from.
+     */
+    public static function record_fuzzy_counts(profile $profile) {
+
+        // Get the profile group record, creating a new one if one does not yet exist.
+        $month = userdate(time() - 360, '%Y%m'); // YYYYMM format.
+        $profilegroup = self::get_record(['name' => $profile->get('groupby'), 'month' => $month]);
```
brendanheywood: Ultimately I want this also in MUC so we don't hit the db at all but save implementing this until right at the end when we know its all working correctly

classes/manager.php
```diff
@@ -243,4 +245,18 @@ public function is_considered_slow(profile $profile): bool {
...
+
+    /**
+     * Increments a counter using the approximate counting algorithm.
+     * Can handle up to PHP_INT_SIZE_BITS counts (2^PHP_INT_SIZE_BITS events).
+     *
+     * @param int $current The current count.
+     * @return int The new count.
+     */
+    public static function approximate_increment(int $current): int {
+        // If the nubmer of events is ever expected to be more than 2 billion, a refactor may be needed.
```
brendanheywood: typo nubmer

classes/manager.php
```diff
@@ -243,4 +245,18 @@ public function is_considered_slow(profile $profile): bool {
...
+        $bits = random_int(PHP_INT_MIN, PHP_INT_MAX);
+        return (($bits << (self::PHP_INT_SIZE_BITS - $current)) == 0) ? $current + 1 : $current;
```
brendanheywood: this is the crux of the whole thing so I think worth spelling it out as an if instead of a ternary and having a comment here with link off to the wikipedia page

classes/page_group.php
brendanheywood: sorry I've misunderstood this code, this looks good to me

classes/page_group.php
```diff
@@ -0,0 +1,114 @@
...
+        // Fuzzy increment the count.
+        $current = $profilegroup->get('fuzzycount');
+        $new = manager::approximate_increment($current);
+        $profilegroup->set('fuzzycount', $new);
+
+        // Fuzzy increment the duration count.
+        $duration = $profile->get('duration');
+        $fuzzydurationcounts = $profilegroup->get('fuzzydurationcounts');
+        $fuzzydurationsum = $profilegroup->get('fuzzydurationsum');
+        $high = 1;
```
brendanheywood: this needs a comment to help grok

I have a feeling something isn't right with the fuzzydurationsum but I'll test it a bit more later and see.

Summing rather than counting will need to tweak the algorithm slightly, if you count 1 hit then you do the random logic once, but if you are summing say a value of 'this page took 10 seconds' then it would be equivalent to running a count of +1 but in a loop doing it 10 times so you have 10 times the chance of bumping the count up a factor of 2.

There is probably a more efficient way of probabilistic summing which doesn't involve a dumb loop. Some out loud thinking:

If the current sum is 2 ^ 4 = 8 and you want to add 8 then we know that 100% of the time it should bump up to to 2 ^ 5 = 16

If the current sum is 2 ^ 4 = 8 and you want to add 4 then we know that 50% of the time it should bump up to to 2 ^ 5 = 16

If the current sum is 2 ^ 4 = 8 and you want to add 2 then we know that 25% of the time it should bump up to to 2 ^ 5 = 16

If the current sum is 2 ^ 4 = 8 and you want to add 1024 then we know that 100% of the time we should practically ignore the current value and bump it directly to 2 ^ 10 = 1024

If the current sum is 2 ^ 9 = 512 and you want to add 1024 then we know that 50% of the time we should bump 2 ^ 10 = 1024 and the other 50% it should be one higher to 2 ^ 11 = 2048

Adding n to m, should be the same as adding m to n (which sounds dumb saying out loud but worth being explicit here)

So I think we need a second algorithm helper function with takes 2 numbers like 8 and 1024 (which may not be factors of 2) and returns at minimum the maximum of the log of both numbers, and then some variable chance of an extra factor bump depending on how close they are.

I don't really understand the current code so maybe it is equivalent to this but not sure

jaypha: I presume you mean 2^4 = 8 and 2^10 = 1024?

brendanheywood: ah github mangled my comments, yes re-read that

jaypha: > I don't really understand the current code so maybe it is equivalent to this but not sure

It does nothing like it, but thanks for clarifying things.

classes/web_processor.php
jaypha: record_fuzzy_counts() is called in one place in web_processor, and in one place in cron_processor. These classes are not active at the same time, so record_fuzzy_counts() should be called only once per run.

brendanheywood: ah sorry yes, so to be consistent should this web one be inside process?

jaypha: Oops, I was mistaken a little there. Let's see if I can explain it better.

`web_processor` creates one profile for the run. `process()` is called multiple times for the same profile in order to make partial saves. Fuzzy counting is done after the final `process()` call. 

`cron_processor` creates a profile for each task it encounters. `process()` is called once for each profile. So fuzzy counting is done each time.

Either way, we do one fuzzy count for each profile attempt. Is this correct?

I moved the call to fuzzy counting out of `cron_processor::process()` anyway.

classes/task/purge_profile_groups.php
jaypha: done

classes/manager.php
jaypha: done

classes/manager.php
jaypha: done

classes/page_group.php
jaypha: I found this a little too vague on what to do when the difference between the logs is 2 or greater, so I just put in a dumb loop.

classes/web_processor.php
brendanheywood: all good

PR #268, opened 2022-09-13
Issue #267: Fix delete profile problems

### LINKED ISSUE #267: Fix deletes with locked profiles
The delete buttons are inconsistent, in some cases deletes don't work. In all cases deletes should never delete a locked profile, not even when doing an individual delete, either we force them to unlock first, or we add a confirmation prompt specifically around it being locked.

There might also be some cross db issues here, I tweaked this logic the other day and I thought I got it working but it isn't in a real site anymore and 'Delete all's are not working. 

Also tweak the land packs to mention locked ones will not be deleted

jaypha:
Resolves #267 

Update delete calls to use $DB->sql_isempty().
Tweak Language strings.
Disable profile delete button when locked.


jaypha:
I tested in on both postgress and mysql.

brendanheywood:
Tested this and looks good to me

Needs a version bump which can come after

PR #281, opened 2022-09-19
Issue #276: Add setting to enable/disable fuzzy counting.

### LINKED ISSUE #276: Have an option to turn the fuzzy count off
Make it default on though, just gives us an option while it is getting ironed out

Make thus a sub option under the main enable, so both need to be on for this to work

jaypha:
Resolves #276

### DIFF
--- a/classes/page_group.php
+++ b/classes/page_group.php
@@ -125,6 +125,11 @@ class page_group extends persistent {
      * @param profile $profile The profile to pull the information from.
      */
     public static function record_fuzzy_counts(profile $profile) {
+        // Do this only if both auto profiling and fuzzy counting is set.
+        if (!get_config('tool_excimer', 'enable_auto') ||
+            !get_config('tool_excimer', 'enable_fuzzy_count')) {
+            return;
+        }
 
         // Get the profile group record, creating a new one if one does not yet exist.
         $month = userdate(time() - 360, '%Y%m'); // YYYYMM format.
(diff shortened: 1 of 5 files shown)


classes/page_group.php
```diff
@@ -61,6 +61,12 @@ protected function set_fuzzydurationcounts(array $data) {
      */
     public static function record_fuzzy_counts(profile $profile) {
```
keevan: Can we remove the extra whitespace here.

classes/page_group.php
```diff
@@ -61,6 +61,12 @@ protected function set_fuzzydurationcounts(array $data) {
...
 
+        // Do this only if both auto profiling and fuzzy counting is set.
+        if (!get_config('tool_excimer', 'enable_auto') ||
+            !get_config('tool_excimer', 'enable_fuzzy_count')) {
```
keevan, line 130: This seems fine. I'm wondering if there are situations where you would only want to put the flag for enable_fuzzy_count here instead, e.g. you might want this to compute the fuzzy count for profiles where it isn't automatically profiled. An example might be a manually triggered cron run, that might have a flag set to profile that particular run.

What do you think?

lang/en/tool_excimer.php
```diff
@@ -52,6 +52,8 @@
 $string['auto_settings_desc'] = 'Settings related to automatic profiling.';
 $string['enable_auto'] = 'Enable auto profiling';
 $string['enable_auto_desc'] = 'Any page will be automatically profiled if they exceed the miniumum duration.';
+$string['enable_fuzzy_count'] = 'Enable fuzzy counting';
+$string['enable_fuzzy_count_desc'] = 'This will cause the plugin to maintain an approximate count of page runs using the {$a}. Automatic profiling must also be enabled.';
```
keevan, line 56: Since lang strings are being added here, could we also bump the version.php?

classes/page_group.php
jaypha: This was explicitly requested.

classes/page_group.php
jaypha: done

lang/en/tool_excimer.php
jaypha: done

PR #282, opened 2022-09-19
Issue #279: Add caching to page groups.

### LINKED ISSUE #279: Cache db calls related to groups
image

jaypha:
Resolves #279

### DIFF
--- a/classes/page_group.php
+++ b/classes/page_group.php
@@ -31,6 +31,40 @@ class page_group extends persistent {
     /** The name of the database table. */
     public const TABLE = 'tool_excimer_page_groups';
 
+    /** Name of the cache used in this class. */
+    public const CACHE_NAME = 'page_group_metadata';
+
+    /**
+     * Creates a page_group, either from the cache, or fresh.
+     *
+     * @param string $name
+     * @param string $month
+     * @return page_group
+     */
+    public static function get_page_group(string $name, string $month): page_group {
+        // Get the cache and attempt to pull the pag group's record from it.
+        $keydata = ['name' => $name, 'month' => $month];
+        $cachekey = serialize($keydata);
+        $cache = \cache::make('tool_excimer', self::CACHE_NAME);
+        $record = $cache->get($cachekey);
+
+        // The cached record exists, return it.
+        if ($record !== false) {
+            return new page_group(0, $record);
+        }
+
+        // Check if a matching DB record exists. If not, create it.
+        $pagegroup = self::get_record($keydata);
+        if ($pagegroup === false) {
+            // No need to set the cache as it is guaranteed to be set later. So return early.
+            return new page_group(0, (object) $keydata);
+        }
+
+        // The DB record should be cached for future calls.
+        $cache->set($cachekey, $pagegroup->to_record());
+        return $pagegroup;
+    }
+
     /**
      * Returns the fuzzy counts for duration.
      *
@@ -55,6 +89,37 @@ class page_group extends persistent {
     }
 
     /**
+     * Called after the page group is created in the DB.
+     */
+    public function after_create() {
+        $this->update_cache();
+    }
+
+    /**
+     * Called after the DB is updated.
+     *
+     * @param bool $result Whether or not the update was successful.
+     */
+    public function after_update($result) {
+        if ($result) {
+            $this->update_cache();
+        }
+    }
+
+    /**
+     * Updates the cache with the current record.
+     */
+    protected function update_cache() {
+        $keydata = [
+            'name' => $this->get('name'),
+            'month' => $this->get('month'),
+        ];
+        $cachekey = serialize($keydata);
+        $cache = \cache::make('tool_excimer', self::CACHE_NAME);
+        $cache->set($cachekey, $this->to_record());
+    }
+
+    /**
      * Records fuzzy count metadata about a page group.
      *
      * @param profile $profile The profile to pull the information from.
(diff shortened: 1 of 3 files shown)


classes/page_group.php
```diff
@@ -54,6 +83,27 @@ protected function set_fuzzydurationcounts(array $data) {
...
+     * Called after the page group is created in the DB.
+     */
+    public function after_create() {
+        $this->after_update(true);
+    }
+
+    /**
+     * Called after the DB is updated.
+     *
+     * @param bool $result
```
keevan: Can we use the parent's phpdoc for the param here, as it is a bit more descriptive.
```suggestion
     * @param bool $result Whether or not the update was successful.
```

classes/page_group.php
```diff
@@ -54,6 +83,27 @@ protected function set_fuzzydurationcounts(array $data) {
         $this->raw_set('fuzzydurationcounts', json_encode($data, JSON_FORCE_OBJECT));
     }
 
+    /**
+     * Called after the page group is created in the DB.
+     */
+    public function after_create() {
+        $this->after_update(true);
```
keevan: This feels like an antipattern. Could we instead call the action we're expecting it to perform as part of creation?

For example:
```
$this->update_cache();
```

classes/page_group.php
```diff
@@ -54,6 +83,27 @@ protected function set_fuzzydurationcounts(array $data) {
...
+     * Called after the DB is updated.
+     *
+     * @param bool $result
+     */
+    public function after_update($result) {
+        if ($result) {
+            $keydata = ['name' => $this->get('name'), 'month' => $this->get('month')];
+            $cachekey = serialize($keydata);
+            $cache = \cache::make('tool_excimer', self::CACHE_NAME);
+            $cache->set($cachekey, $this->to_record());
```
keevan: Similar to an above comment, could we pull out this block into an internal method and call it when needed. I was going to ask for a comment to be added here but this seems like a better approach given the need to reuse it.

For example:
```
$this->update_cache();
```

classes/page_group.php
```diff
@@ -31,6 +31,35 @@ class page_group extends persistent {
...
+     *
+     * @param string $name
+     * @param string $month
+     * @return page_group
+     */
+    public static function get_page_group(string $name, string $month): page_group {
+        $keydata = ['name' => $name, 'month' => $month];
+        $cachekey = serialize($keydata);
+        $cache = \cache::make('tool_excimer', self::CACHE_NAME);
+        $record = $cache->get($cachekey);
```
keevan, line 49: Could we add a comment to this block, perhaps something related to it attempting to pull out the data from cache first.

classes/page_group.php
```diff
@@ -31,6 +31,35 @@ class page_group extends persistent {
...
+            $pagegroup = self::get_record($keydata);
+            if ($pagegroup === false) {
+                // No need to set the cache as it is guaranteed to be set later.
+                return new page_group(0, (object) $keydata);
+            }
+            $cache->set($cachekey, $pagegroup->to_record());
+            return $pagegroup;
+        } else {
+            return new page_group(0, $record);
+        }
```
keevan: I found this slightly difficult to follow, so I'm going to lay out what I'm reading.

If the record wasn't cached, then get the matching record from the DB based on the key data.

If the record wasn't found from the previous, create a new page_group record containing just the key data (month, name). This should default the other fields like fuzzycount to zero, so those don't need to be explicitly set from what it appears. This returns early as no extra cache set step needs to happen at this point.

If the DB record does exist, it caches it and returns the current record (same as cached)

Otherwise, if the cached record existed originally, it will return a page_group object matching the cached record.

--

Could this can be reordered and simplified? Here's an example below:
```suggestion
        // The cached record exists, return it.
        if ($record !== false) {
            return new page_group(0, $record);
        }

        // Check if a matching DB record exists. If not, create it.
        $pagegroup = self::get_record($keydata);
        if ($pagegroup === false) {
            // No need to set the cache as it is guaranteed to be set later. So return early.
            return new page_group(0, (object) $keydata);
        }

        // The DB record should be cached for future calls.
        $cache->set($cachekey, $pagegroup->to_record());
        return $pagegroup;
```

keevan (commented):
Hey @jaypha, it looks mostly good. I've left some comments below. 

As db/caches.php has been modified, can we also bump version.php?

classes/page_group.php
```diff
@@ -31,6 +31,41 @@ class page_group extends persistent {
...
+
+    /**
+     * Creates a page_group, either from the cache, or fresh.
+     *
+     * @param string $name
+     * @param string $month
+     * @return page_group
+     */
+    public static function get_page_group(string $name, string $month): page_group {
+
```
keevan: Can we remove the extra whitespace here please.

classes/page_group.php
```diff
@@ -54,6 +89,36 @@ protected function set_fuzzydurationcounts(array $data) {
...
+    public function after_update($result) {
+        if ($result) {
+            $this->update_cache();
+        }
+    }
+
+    /**
+     * Updates the cache with the current record.
+     *
+     * @throws \coding_exception
```
keevan: This method/function doesn't seem like it should have a @throws tag. If one was required, I expect to see one explicitly thrown from this method/function

Reference: https://docs.moodle.org/dev/Coding_style#.40throws

classes/page_group.php
```diff
@@ -54,6 +89,36 @@ protected function set_fuzzydurationcounts(array $data) {
...
+     */
+    protected function update_cache() {
+        $keydata = ['name' => $this->get('name'), 'month' => $this->get('month')];
```
keevan: Small thing - which is a personal opinion. It may be slightly clearer to see the key values of this multi-value array if they were put on multiple lines. The reasoning behind the suggestion is because the values aren't simple variables, and instead are calls to methods with params.

For example:
```suggestion
        $keydata = [
            'name' => $this->get('name'), 
            'month' => $this->get('month'),
        ];
```

PR #283, opened 2022-09-19
Issue #278: Remove use of get_admin().

### LINKED ISSUE #278: get_admin() does a db request when all we need is the admin id which we already have

jaypha:
Resolves #278

### DIFF
--- a/classes/profile.php
+++ b/classes/profile.php
@@ -268,17 +268,22 @@ class profile extends persistent {
             return true;
         }
 
-        // If current userid isn't guest, but stored user is guest, update it.
+        // If the would-be user matches the currently stored, do not update.
+        if ($currentid === $stored) {
+            return false;
+        }
+
+        // If stored user is guest, update it.
         $guestid = (int) $CFG->siteguest;
-        if ($guestid && $currentid !== $guestid && $stored === $guestid) {
+        if ($guestid && $stored === $guestid) {
             return true;
         }
 
-        // If current user isn't cli admin, but stored user is cli admin, update it.
-        $adminid = get_admin()->id;
-        if ($adminid && $currentid !== $adminid && $stored === $adminid) {
+        // If stored user is cli admin, update it.
+        if (!empty($CFG->siteadmins) && in_array($stored, explode(',', $CFG->siteadmins))) {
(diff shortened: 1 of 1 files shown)


classes/profile.php
```diff
@@ -268,17 +268,21 @@ protected function check_update_userid(int $currentid) : bool {
             return true;
         }
 
-        // If current userid isn't guest, but stored user is guest, update it.
+        if ($currentid === $stored) {
+            return false;
+        }
```
keevan, line 274: This block is great. You've managed to simplify the following blocks by pulling out a common factor and returning a result early, and also because of that it should in most cases be quicker to evaluate which means better performance. :gem: 

Could we add a comment to this block? It would make it more consistent with the rest of the conditional blocks in this method and reinforces the intent. Something along the lines of "If the would-be user matches the currently stored, do not update".

classes/profile.php
```diff
@@ -268,17 +268,21 @@ protected function check_update_userid(int $currentid) : bool {
...
-        if ($guestid && $currentid !== $guestid && $stored === $guestid) {
+        if ($guestid && $stored === $guestid) {
             return true;
         }
 
-        // If current user isn't cli admin, but stored user is cli admin, update it.
-        $adminid = get_admin()->id;
-        if ($adminid && $currentid !== $adminid && $stored === $adminid) {
+        // If stored user is cli admin, update it.
+        if (!empty($CFG->siteadmins) && in_array($stored, explode(',', $CFG->siteadmins))) {
```
keevan, line 283: Since the stored value will always (by this stage) be an _int_. Could we simplify this further and based on current experience, assume there is always going to be a siteadmin value of some sort set (as it would be most of the time), and go straight for the explode?

For example:
```suggestion
        if (in_array($stored, explode(',', $CFG->siteadmins))) {
```

If siteadmins could be ever be null, then perhaps it can stay in its current form.

jaypha: According to the comments inside `get_admin()`, siteadmins can be null during unit tests.

classes/profile.php
jaypha: done

PR #285, opened 2022-09-20
Issue #272: Ensure that indexes are grouped together.

### LINKED ISSUE #272: Fuzzy counts work for /index.php vs / are not grouped together
We should fix this at the grouping level, any urls which are the same like / and /index.php should be merged into the same group, probably without the index.php

Special case, show the / at the root level as '/' and not as an empty string

jaypha:
Resolves #272

### DIFF
--- a/classes/script_metadata.php
+++ b/classes/script_metadata.php
@@ -145,24 +145,15 @@ class script_metadata {
     }
 
     /**
-     * Gets the name of the script.
+     * Gets the name of the script that is being run for the request.
      *
      * @return string the request path for this profile.
      */
-    public static function get_request(): string {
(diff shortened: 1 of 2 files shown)


brendanheywood:
I’d like to see some tests for this with data provider covering cases like a url which doesn’t exist and got to an error page. I have a spidey sense using SCRIPT may break some things

jaypha:
This is being put into the 'too hard' basket for now. The question being how can we determine if / and /index.php are referring to the same script, especially if grouping by request name.

brendanheywood:
This was done in a another pr https://github.com/catalyst/moodle-tool_excimer/pull/328

PR #286, opened 2022-09-21
Issue #280: Add extra redact keys.

### LINKED ISSUE #280: Redact web service tokens in urls
* [ ] POST webservice/rest/server.php?wstoken=xxxxxxxx&wsfunction=

Have a scan through some prod sites access logs and look for other things we should redact

* [ ] ?key=xx
* [ ] ?token=xxx

jaypha:
Resolves #280

### DIFF
--- a/classes/script_metadata.php
+++ b/classes/script_metadata.php
@@ -34,29 +34,35 @@ class script_metadata {
 
 
     /** List of parameters that are not to be recorded at all. */
-    const DENYLIST = [
+    const DENY_LIST = [
         manager::FLAME_ME_PARAM_NAME,
-        manager::FLAME_ON_PARAM_NAME,
         manager::FLAME_OFF_PARAM_NAME,
+        manager::FLAME_ON_PARAM_NAME,
     ];
 
-    /** List of paramteres that are to be recorded in redacted form. */
-    const REDACTLIST = ['sesskey'];
+    /** List of parameters that are to be recorded in redacted form. */
+    const REDACT_LIST = [
+        'authtoken',
+        'key',
+        'nonce',
+        'sesskey',
+        'wstoken',
+    ];
 
     /**
      * List of script names that requires more info for grouping.
      * TODO: This list is incomplete.
      */
     const SCRIPT_NAMES_FOR_GROUP_REFINING = [
+        'admin/category.php',
         'admin/index.php',
         'admin/settings.php',
         'admin/search.php',
-        'admin/category.php',
-        'lib/javascript.php',
         'lib/ajax/service.php',
+        'lib/javascript.php',
         'pluginfile.php',
-        'webservice/pluginfile.php',
         'tokenpluginfile.php',
+        'webservice/pluginfile.php',
     ];
 
     /** Minimum sampling period. */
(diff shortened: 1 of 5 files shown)


settings.php
```diff
@@ -128,6 +128,26 @@
             )
         );
 
+        // Get a list of builtin redactable params, quoted and separated by commas.
+        $builtin = implode(
+            ', ',
+            array_map(
+                function($v) {
```
keevan: There should be one space after the function keyword for anonymous functions.

classes/script_metadata.php
```diff
@@ -40,8 +40,14 @@ class script_metadata {
         manager::FLAME_OFF_PARAM_NAME,
     ];
 
-    /** List of paramteres that are to be recorded in redacted form. */
-    const REDACTLIST = ['sesskey'];
+    /** List of parameters that are to be recorded in redacted form. */
+    const REDACTLIST = [
```
keevan: Constants should typically be in uppercase and separated by underscores for each word used.

In this case, it would probably be REDACT_LIST

https://docs.moodle.org/dev/Coding_style#Constants_2

classes/script_metadata.php
```diff
@@ -40,8 +40,14 @@ class script_metadata {
...
+        'sesskey',
+        'wstoken',
+        'authtoken',
+        'key',
+        'nonce',
```
keevan: Could we sort this list? It would be quicker to look things up if things were alphabetical.

classes/script_metadata.php
```diff
@@ -136,14 +142,43 @@ function ($i) {
...
+
+        $lines = explode(PHP_EOL, $setting);
+
+        $params = self::REDACTLIST;
+        foreach ($lines as $line) {
+            // Strip # comments, and trim.
+            $line = trim(preg_replace('/#.*$/', '', $line));
+
+            // Ignore empty lines.
+            if ($line == '') {
```
keevan: Could we use `===` here instead of `==`?

classes/script_metadata.php
```diff
@@ -162,8 +197,6 @@ public static function get_request(): string {
         $request = str_replace($CFG->wwwroot, '', $request);
         $request = ltrim($request, '/');
         return $request;
-
```
keevan, line 165: Nice catch here :)

classes/script_metadata.php
```diff
@@ -136,14 +142,43 @@ function ($i) {
...
+    public static function get_redactable_param_names(): array {
+        $setting = get_config('tool_excimer', 'redact_params');
+
+        // Strip C style comments.
+        $setting = preg_replace('!/\*.*?\*/!s', '', $setting);
+
+        $lines = explode(PHP_EOL, $setting);
+
+        $params = self::REDACTLIST;
+        foreach ($lines as $line) {
```
keevan, line 169: A block comment explaining the purpose of this loop would be good. What do you think?

keevan (commented):
Tests look good, there are a few suggestions I've added below.

Looks okay to merge after that though. Thanks @jaypha

classes/script_metadata.php
jaypha: The comment on line 167 addresses the loop. The loop is also self-explanetary, so a comment shouldn't be needed.

keevan: Thanks for adding a comment. Also I realise that "block comment" might have refered to a larger comment. What I actually intended was a comment on the code block, which you have added.

PR #289, opened 2022-09-22
Issue #274: Display page groups by month.

### LINKED ISSUE #274: Fuzzy counts show only the current month by default
Have a filter to go back a month, probably do this as a pagination bar above the table

jaypha:
Resolves #274 

Add month selector to page group table page.

### DIFF
--- a/classes/output/renderer.php
+++ b/classes/output/renderer.php
@@ -31,8 +31,6 @@ class renderer extends \plugin_renderer_base {
      *
      * @param tabs $tabs
      * @return string
-     * @throws \coding_exception
-     * @throws \moodle_exception
      */
     public function render_tabs(tabs $tabs): string {
         $selected = '';
@@ -46,5 +44,15 @@ class renderer extends \plugin_renderer_base {
             $tabobjects[] = $tmp;
         }
         return $this->output->tabtree($tabobjects, $selected);
+    }
+
+    /**
+     * Renders the month selector utility.
+     *
+     * @param array $data Data to be passed to the template.
+     * @return bool|string
+     */
+    public function render_month_selector(array $data) {
+        return $this->render_from_template('tool_excimer/month_selector', $data);
     }
 }
--- a/classes/page_group.php
+++ b/classes/page_group.php
@@ -35,6 +35,25 @@ class page_group extends persistent {
     public const CACHE_NAME = 'page_group_metadata';
 
     /**
+     * Returns the current month in YYYYMM format.
+     * @return int
+     */
+    public static function get_current_month(): int {
+        return (int) userdate(time(), '%Y%m');
+    }
+
+    /**
+     * Gets the number of records stored for a paricular month.
+     *
+     * @param int $month
+     * @return bool
+     */
+    public static function record_exists_for_month(int $month): bool {
+        global $DB;
+        return $DB->record_exists(self::TABLE, ['month' => $month]);
+    }
+
+    /**
      * Creates a page_group, either from the cache, or fresh.
      *
      * @param string $name
@@ -132,7 +151,7 @@ class page_group extends persistent {
         }
 
         // Get the profile group record, creating a new one if one does not yet exist.
-        $month = userdate(time() - 360, '%Y%m'); // YYYYMM format.
+        $month = self::get_current_month();
         $pagegroup = self::get_page_group($profile->get('groupby'), $month);
 
         $existing = $pagegroup->to_record();
--- a/classes/page_group_table.php
+++ b/classes/page_group_table.php
@@ -29,11 +29,13 @@ class page_group_table extends \table_sql {
     /** Columns to be displayed. */
     const COLUMNS = [
         'name',
-        'month',
         'fuzzycount',
         'fuzzydurationcounts',
         'fuzzydurationsum',
     ];
+
+    /** @var int The month to display in YYYYMM format. */
+    private $month = null;
 
     /**
      * Defines the columns for this table.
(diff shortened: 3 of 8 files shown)


classes/output/renderer.php
```diff
@@ -47,4 +45,14 @@ public function render_tabs(tabs $tabs): string {
         }
         return $this->output->tabtree($tabobjects, $selected);
     }
+
+    /**
+     * Renders the month selector utility.
+     *
+     * @param array $data Dat to be passed to the template.
```
keevan: Dat -> Data

classes/page_group_table.php
```diff
@@ -71,7 +82,8 @@ public function setup() {
         $this->set_sql(
             '*',
             '{' . page_group::TABLE . '}',
-            '1=1'
+            $this->month ? 'month = ?' : '1=1',
+            $this->month ? [$this->month] : []
```
keevan: Can we move this into a previous prepare block and do something if the month is set, and do something else if the month is not set? This would keep it more consistent with wider usage and will be easier to read.

e.g.
```
// Defaults ...

// When month is set.
if ($this->month) {
    ...
}

$this->set_sql(
    '*', 
    '{' . page_group::TABLE . '}',
    $where,
    $params
);
```

classes/page_group.php
```diff
@@ -34,6 +34,25 @@ class page_group extends persistent {
     /** Name of the cache used in this class. */
     public const CACHE_NAME = 'page_group_metadata';
 
+    /**
+     * Returns the current month in YYYYMM format.
+     * @return int
+     */
+    public static function get_current_month(): int {
+        return (int) userdate(time() - 360, '%Y%m');
```
keevan: Can we add a description and clarification for what the 360 is used for here?

classes/page_group_table.php
```diff
@@ -29,12 +29,14 @@ class page_group_table extends \table_sql {
     /** Columns to be displayed. */
     const COLUMNS = [
         'name',
-        'month',
         'fuzzycount',
         'fuzzydurationcounts',
         'fuzzydurationsum',
     ];
 
+    /** @var int The month to display in YYYYMM format (will display current month if null). */
```
keevan: Based on the setup, is the month ever going to be null? Which might mean the later part of this description may need to change.

classes/page_group_table.php
keevan: Also to reiterate the previous comment, would $this->month ever be null (in practice)?

page_groups.php
```diff
@@ -59,6 +64,31 @@
...
+    }
+
+    $prevmonth = $month - 1;
+    if ($prevmonth % 100 == 0) {
+        $prevmonth -= 88; // Subtract 1 year, but add 12 months.
+    }
+    $nextmonth = $month + 1;
+    if ($nextmonth % 100 == 13) {
+        $nextmonth += 88; // Add one year, but subract 12 months.
+    }
```
keevan: There are a few magic numbers in the mix here. Could we add a description or set them to a variable to make them more explanatory?

classes/page_group.php
```diff
@@ -34,6 +34,25 @@ class page_group extends persistent {
...
+
+    /**
+     * Gets the number of records stored for a paricular month.
+     *
+     * @param int $month
+     * @return int
+     */
+    public static function count_records_for_month(int $month): int {
+        global $DB;
+        return $DB->count_records(self::TABLE, ['month' => $month]);
```
keevan: Since the `count_records_for_month` function is only used to check for emptyness/existence for nexturl/prevurl in the `page_groups.php` file. Could we instead change the function to perform a `record_exists` check instead, which should be faster theoretically as it does not need to compute the sum.

page_groups.php
keevan: Could we update these to be strict comparisons, e.g.  `===` instead of `==`

keevan:
Also, after some use, I'm wondering if it's better to display a list of months running back, to make it easier to click to. And perhaps a drop down for anything further back than that. Currently you would need to go back month by month. 

image

Though full blown pagination might be better suited for things stored for a longer period of time, excimer might not be the place for this. (Loosely thinking about the cleanup script)

What are your thoughts?

classes/output/renderer.php
jaypha: done

classes/page_group_table.php
jaypha: I've refactored that there will always be a month set.

classes/page_group.php
jaypha: I frankly don't know what it is for myself. I got rid of it.

classes/page_group_table.php
jaypha: See above.

page_groups.php
jaypha: I have added some comments, but will am refactoring this further as part of the other PR.

classes/page_group.php
jaypha: done

jaypha:
> What are your thoughts?
It has merit, but I need to ask that this be put into another issue. I am running out of time.

classes/page_group_table.php
```diff
@@ -63,15 +65,29 @@ protected function get_columns(): array {
...
+
+    /**
+     * Overrides flexible_table::setup() to do some extra setup.
      *
      * @return false|\type|void
+     * @throws \moodle_exception
      */
     public function setup() {
+        if (is_null($this->month)) {
+            throw new \moodle_exception('*Month value not set in page group table.');
```
keevan: Could we change this into a lang string?

keevan (commented):
A tiny comment, but otherwise looks good. 

Thanks @jaypha

classes/page_group_table.php
jaypha: Is there a 'pull my hair out ' emoji?

jaypha: I even put a little asterix to remind me. Although now that I think about it it seems rather useless.

keevan: haha I understand. It feels like there needs to be some sort of phpcs sniff to catch this. I've hit this a lot so much so I force myself (though admittedly sometimes forgetting) to "just do it".

PR #291, opened 2022-09-23
Issue #288: Format numbers displayed on profile page.

### LINKED ISSUE #288: Improve the flame graph durations to be human readable durations
image

* [ ] Display this as hours / minutes / seconds. If it is seconds then also show 3 decimals ie milliseconds but don't show that if we are in the realm of minutes or bigger
* [ ] also for any other numbers like the samples, run this through format_float to add the thousands separator

jaypha:
Resolves #288

### DIFF
--- a/profile.php
+++ b/profile.php
@@ -153,9 +153,22 @@ $data['delete_all_button'] = $output->render($deleteallbutton);
 $data['profile_lock_button'] = $output->render($lockprofilebutton);
 
 $data['responsecode'] = helper::status_display($profile->get('scripttype'), $profile->get('responsecode'));
-// Totara doesn't like the mustache string helper being called with varaibles.
-$data['numsamples_str'] = get_string('field_numsamples_value', 'tool_excimer',
-    ['samples' => $data['numsamples'], 'samplerate' => $data['samplerate']]);
+
+$decsep = get_string('decsep', 'langconfig');
+$thousandssep = get_string('thousandssep', 'langconfig');
+
+// Totara doesn't like the mustache string helper being called with variables.
+$data['numsamples_str'] = get_string(
+    'field_numsamples_value',
+    'tool_excimer',
+    [
+        'samples' => number_format($data['numsamples'], 0, $decsep, $thousandssep),
+        'samplerate' => $data['samplerate'],
+    ]
+);
+$data['dbreads'] = number_format($data['dbreads'], 0, $decsep, $thousandssep);
+$data['dbwrites'] = number_format($data['dbwrites'], 0, $decsep, $thousandssep);
+$data['dbreplicareads'] = number_format($data['dbreplicareads'], 0, $decsep, $thousandssep);
 
 if ($user) {
     $data['userlink'] = new moodle_url('/user/profile.php', ['id' => $user->id]);
@@ -168,6 +181,7 @@ $data['lockreason'] = format_text($data['lockreason']);
 $tabs = new tabs($url);
 
 $data['tabs'] = $tabs->export_for_template($output)['tabs'];
+$data['lang'] = get_config('core', 'lang');
 
 echo $output->header();
 echo $output->render_from_template('tool_excimer/profile', $data);
--- a/templates/flamegraph.mustache
+++ b/templates/flamegraph.mustache
@@ -27,10 +27,12 @@
 
     Context variables required for this template:
     * id - The id of the profile
+    * lang - Language setting
 
     Example context (json):
     {
-        "id" : 3
+        "id" : 3,
+        "lang" : "en"
     }
 }}
 
@@ -47,6 +49,34 @@
 
 init();
 window.addEventListener('resize', _ => init());
+
+function displaysamplestats(d) {
+    let time = d.value * {{samplerate}};
+    let seconds = time / 1000;
+    let minutes = Math.floor(seconds / 60);
+    let hours = Math.floor(minutes / 60);
+    minutes = minutes % 60;
+    seconds = (seconds % 60).toFixed(3);
+
+    let durationstr = '';
+    if (hours) {
+        durationstr += `${hours}h `;
+    }
+    if (minutes) {
+        durationstr += `${minutes}m `;
+    }
+    if (seconds) {
+        if (durationstr !== '') {
+            seconds = Math.round(seconds);
+        }
+        durationstr += `${seconds}s`;
+    }
+
+    const percentage = ((d.x1 - d.x0) * 100).toFixed(3);
+    const samplecount = d.value.toLocaleString("{{lang}}");
+
+    return d.data.name + ` (${percentage}%, ${samplecount} {{#str}} samples, tool_excimer {{/str}}, {{#str}} duration, tool_excimer {{/str}} ~ ${durationstr}})`;
+}
 
 function init() {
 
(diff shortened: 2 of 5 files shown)


profile.php
```diff
@@ -153,9 +153,22 @@
 $data['profile_lock_button'] = $output->render($lockprofilebutton);
 
 $data['responsecode'] = helper::status_display($profile->get('scripttype'), $profile->get('responsecode'));
-// Totara doesn't like the mustache string helper being called with varaibles.
-$data['numsamples_str'] = get_string('field_numsamples_value', 'tool_excimer',
-    ['samples' => $data['numsamples'], 'samplerate' => $data['samplerate']]);
+
+$decsep = get_string('decsep', 'langconfig');
```
brendanheywood, line 157: can all of this be done using the moodle format_float api and not the raw php one?

jaypha: Moodle's format_float does not use thousands separators. If this is OK, I will change this.

brendanheywood: Ah yeah I’ve hit that before and hq didn’t want to add it, at least not by default. Ok leave this as is

templates/flamegraph.mustache
```diff
@@ -73,16 +75,22 @@ function init() {
...
-                + '%, ' + d.value + ' samples, duration ~' +  d.value * {{samplerate}} + ' ms)'
-    });
+    displaysamplestats = function (d) {
+        time = d.value * {{samplerate}};
+        seconds = time / 1000;
+        minutes = Math.floor(seconds / 60);
+        hours = Math.floor(minutes / 60);
+        minutes = minutes % 60;
+        seconds = (seconds % 60).toFixed(3);
+        durationstr = (hours ? (hours + 'h ') : '') + (minutes ? (minutes + 'm ') : '') + seconds + 's';
```
keevan: Could this function be moved to the global scope? Alternatively, if that's not possible, it would be better to use `const displaysamplestats` to declare it as it would be scoped properly.

templates/flamegraph.mustache
```diff
@@ -73,16 +75,22 @@ function init() {
         draw();
     }
 
-    window.excimerChart.label(function(d) {
-        return d.data.name + ' (' + ((d.x1 - d.x0) * 100).toFixed(3)
-                + '%, ' + d.value + ' samples, duration ~' +  d.value * {{samplerate}} + ' ms)'
-    });
+    displaysamplestats = function (d) {
```
keevan: According to the issue, it should only display the decimal when things are in the realm of seconds.

I'm currently seeing this locally, and the conditions would need to be added here.

image

templates/flamegraph.mustache
```diff
@@ -73,16 +75,22 @@ function init() {
...
+        time = d.value * {{samplerate}};
+        seconds = time / 1000;
+        minutes = Math.floor(seconds / 60);
+        hours = Math.floor(minutes / 60);
+        minutes = minutes % 60;
+        seconds = (seconds % 60).toFixed(3);
+        durationstr = (hours ? (hours + 'h ') : '') + (minutes ? (minutes + 'm ') : '') + seconds + 's';
```
keevan: This line is a bit hard to read. Are we able to split this out and clean it up a bit. This also applies to the return statement further below. There's a bunch of calculation happening within the same line of code.
```suggestion
        let durationstr = '';
        if (hours) {
            durationstr += `${hours}h `;
        }
        if (minutes) {
            durationstr += `${minutes}m `;
        }
        if (seconds) {
            durationstr += `${seconds}s`;
        }
```

templates/flamegraph.mustache
```diff
@@ -73,16 +75,22 @@ function init() {
...
 
-    var tip = flamegraph.tooltip.defaultFlamegraphTooltip().text(function(d) {
         return d.data.name + ' (' + ((d.x1 - d.x0) * 100).toFixed(3)
-        + '%, ' + d.value + ' samples, duration ~' +  d.value * {{samplerate}} + ' ms)'
-    });
+                + '%, ' + d.value.toLocaleString("{{lang}}") + ' samples, duration ~ ' +  durationstr + ')'
```
keevan: Can we please clean this up to make it easier to read and make future changes?

```suggestion
        const percentage = ((d.x1 - d.x0) * 100).toFixed(3)
        const sampleCount = d.value.toLocaleString("{{lang}}")
        return d.data.name + ` (${percentage}%, ${sampleCount} samples, duration ~ ${durationstr})`
```

templates/flamegraph.mustache
```diff
@@ -73,16 +75,22 @@ function init() {
...
+    }
+
+    window.excimerChart.label(displaysamplestats);
 
+    var tip = flamegraph.tooltip.defaultFlamegraphTooltip().text(displaysamplestats);
```
keevan: Small cleanup here would be good at the same time.

`var` -> `const`

templates/memoryusagegraph.mustache
```diff
@@ -182,14 +182,14 @@ function processGraph(id, data) {
...
-              Sample #${d.sampleindex} <br/>
+              Sample ${d.sampleindex.toLocaleString('{{lang}}')} <br/>
               Memory: ${display_size(d.value)}
           `)
               .style("left", (event.pageX) + "px")
               .style("top", (event.pageY - 28) + "px");
 
           memoryusagedetails.innerHTML = `
-              Sample #${d.sampleindex}, Memory Usage: ${display_size(d.value)}
+              Sample ${d.sampleindex.toLocaleString('{{lang}}')}, Memory Usage: ${display_size(d.value)}
```
keevan: As it's starting to do more, could we move this into a local `const` and share its usage?

keevan (commented):
Thanks Jason, results look much better. 

Some small adjustments needed and it would seem good to merge in.

templates/flamegraph.mustache
keevan: It's recommended that we declare variables with `let` or `const` instead of leaving them in the global scope which is the default (See https://stackoverflow.com/a/60610552/4757871).

`let` would be for variables where you expect them to be reassigned / changed.
`const` would be values for which you would not change after setting them initially (for objects this would mean their reference)

jaypha: done

templates/flamegraph.mustache
jaypha: done

templates/flamegraph.mustache
jaypha: done

templates/flamegraph.mustache
jaypha: done

templates/flamegraph.mustache
jaypha: done

templates/memoryusagegraph.mustache
jaypha: done

keevan (commented):
Changes look great! Thanks for applying the suggestions @jaypha 

Could we bump the version as the lang strings file was also modified. 

I'm going to split out the remaining "string not in lang strings" issue out into a separate task as you've made me aware of the ones e.g. in memoryusagegraph.mustache (here https://github.com/catalyst/moodle-tool_excimer/issues/293)

PR #292, opened 2022-09-26
Issue #275: Create a separate page for each page group.

### LINKED ISSUE #275: Fuzzy count, drill down into a specific url group
* [x] Show the histogram of counts and times for that page
* [x] show the trends of the histogram across the last N months

jaypha:
Resolves #275 

Add table to display attributes for page group.
Add chart to display historical data for histogram.

### DIFF
--- a/classes/helper.php
+++ b/classes/helper.php
@@ -183,4 +183,58 @@ class helper {
         }
         return $displayedrequest;
     }
+
+    /**
+     * Make a single record for a histogram table.
+     * Row is of the form: 2^(k-1) - 2^k : 2^(v-1).
+     *
+     * @param int $durationexponent The exponent (k) of the high end of the duration range.
+     * @param int $count The fuzzy count (v), or zero if no value.
+     * @return array
+     */
+    private static function make_histogram_record(int $durationexponent, int $count): array {
+        $high = pow(2, $durationexponent);
+        $low = ($high === 1) ? 0 : pow(2, $durationexponent - 1);
+        $val = $count ? pow(2, $count - 1) : 0;
+        return [
+            'low'   => $low,
+            'high'  => $high,
+            'value' => $val,
+        ];
+    }
+
+    /**
+     * Create a histogram table for a page group.
+     *
+     * @param \stdClass $record Page group's record.
+     * @return array
+     */
+    public static function make_histogram(\stdClass $record): array {
+        $counts = json_decode($record->fuzzydurationcounts, true);
+        ksort($counts);
+
+        $histogramrecords = [];
+        $durationexponent = 0;
+        // Generate a line for each duration range up to the highest stored in the DB.
+        foreach ($counts as $storeddurationexponent => $fuzzycount) {
+            // Fill in lines that do not have stored values.
+            while ($durationexponent < $storeddurationexponent) {
+                $histogramrecords[] = self::make_histogram_record($durationexponent, 0);
+                ++$durationexponent;
+            }
+            $histogramrecords[] = self::make_histogram_record($storeddurationexponent, $fuzzycount);
+            ++$durationexponent; // Ensures this line is not printed twice.
+        }
+        return $histogramrecords;
+    }
+
+    /**
+     * Formats a monthint value with the mmm YYYY format. (e.g. Dec 2020).
+     *
+     * @param int $monthint
+     * @return string
+     */
+    public static function monthint_formatted(int $monthint): string {
+        return userdate(monthint::as_timestamp($monthint), get_string('strftime_monyear', 'tool_excimer'));
+    }
 }
--- a/page_group.php
+++ b/page_group.php
@@ -1,0 +1,147 @@
+<?php
+// This file is part of Moodle - https://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Page group meta data.
+ *
+ * @package   tool_excimer
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+use core\chart_line;
+use core\chart_series;
+use tool_excimer\helper;
+use tool_excimer\monthint;
+use tool_excimer\output\tabs;
+use tool_excimer\page_group;
+
+require_once('../../../config.php');
+require_once($CFG->libdir.'/adminlib.php');
+
+$expiry = get_config('tool_excimer', 'expiry_fuzzy_counts');
+$pagegroupid = required_param('id', PARAM_INT);
+$monthstodisplay = optional_param('monthstodisplay', $expiry, PARAM_INT);
+
+$url = new \moodle_url('/admin/tool/excimer/page_group.php');
+$context = context_system::instance();
+
+$pagegroup = new page_group($pagegroupid);
+
+$PAGE->set_context($context);
+$PAGE->set_url($url);
+$PAGE->navbar->add($pagegroup->get('name'));
+
+admin_externalpage_setup('tool_excimer_report_page_groups');
+
+$output = $PAGE->get_renderer('tool_excimer');
+
+$data = $pagegroup->to_record();
+
+// Data for table.
+$data->month = helper::monthint_formatted($data->month);
+$data->approxcount = pow(2, $data->fuzzycount - 1) . ' - ' . pow(2, $data->fuzzycount);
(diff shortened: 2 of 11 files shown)


classes/helper.php
```diff
@@ -183,4 +183,48 @@ public static function full_request(\stdClass $profile): string {
...
+     * Make a single record for a histogram table.
+     * Row is of the form: 2^(k-1) - 2^k : 2^(v-1).
+     *
+     * @param int $durationexponent The exponent (k) of the high end of the duration range.
+     * @param int $count The fuzzy count (v), or zero if no value.
+     * @return array
+     */
+    private static function make_histogram_record(int $durationexponent, int $count): array {
+        $high = pow(2, $durationexponent);
+        $low = ($high == 1) ? 0 : pow(2, $durationexponent - 1);
```
keevan: Can we use strict checks here?
`==` -> `===`

templates/page_group.mustache
```diff
@@ -0,0 +1,82 @@
...
+        </div>
+        <div>
+            <table class="flexible table table-striped table-sm table-hover generaltable generalbox w-auto">
+                <tr>
+                    <th>{{#str}}field_fuzzydurationcounts, tool_excimer{{/str}}</th>
+                    <td></td>
+                </tr>
+                {{#histogram}}
+                    <tr>
+                        <td>{{low}} - {{high}}s</td>
```
keevan, line 75: The low high display on the page_group and page_groups list page are slightly different due to spacing (e.g. here vs the lang string `fuzzydurationrange`)

Could we update to make them consistent?

Page groups:
image

Page group:
image

templates/page_group.mustache
```diff
@@ -0,0 +1,82 @@
...
+        "approxduration" : 212,
+        "histogram" : [ { "low" : 0, "high" : 1, "value" : 3 } ]
+    }
+}}
+
+<h3>{{name}}</h3>
+
+<div id="profile-stats" class="container">
+    <div class="row">
+        <div class="">
```
keevan: We can probably remove the empty `class=""` bit here.

page_group.php
```diff
@@ -0,0 +1,145 @@
...
+
+$PAGE->set_title($pluginname);
+$PAGE->set_pagelayout('admin');
+$PAGE->set_heading($pluginname);
+
+echo $output->header();
+$tabs = new tabs($url);
+echo $output->render_tabs($tabs);
+echo $output->render_from_template('tool_excimer/page_group', $data);
+echo html_writer::tag('h3', 'Histogram history');
```
keevan: Can we convert and use `get_string()` here

> ... to present the best text to every user

https://docs.moodle.org/dev/String_API#Overview

classes/helper.php
```diff
@@ -183,4 +183,48 @@ public static function full_request(\stdClass $profile): string {
...
+     * Create a histogram table for a page group.
+     *
+     * @param \stdClass $record Page group's record.
+     * @return array
+     */
+    public static function make_histogram(\stdClass $record): array {
+        $counts = json_decode($record->fuzzydurationcounts, true);
+        ksort($counts);
+
+        $histogramrecs = [];
```
keevan: Could we use the full name here?

```suggestion
        $histogramrecords = [];
```

page_group.php
```diff
@@ -0,0 +1,145 @@
...
+$PAGE->navbar->add($pagegroup->get('name'));
+
+admin_externalpage_setup('tool_excimer_report_page_groups');
+
+$output = $PAGE->get_renderer('tool_excimer');
+
+$data = $pagegroup->to_record();
+
+// Data for table.
+$data->month = userdate(strtotime($data->month . '01'), '%b %Y');
```
keevan: There does not seem to be an equivalent format as this one, but a similar one exists as `'%B %Y'`

Which would mean the display changes from `Sep 2022` -> `September 2022`. That said I'm a bit reluctant to change it to the long form. It is being used also on the page_group_table class. Could we move this format to a `const` and reference it in both places for consistency?

The alternative would be to move this into a helper function (which might happen for something similar below), perhaps something such as `get_month_formatted(int $timestamp)`

page_group.php
```diff
@@ -0,0 +1,145 @@
...
+$data->approxcount = pow(2, $data->fuzzycount - 1) . ' - ' . pow(2, $data->fuzzycount);
+$data->approxduration = pow(2, $data->fuzzydurationsum);
+$data->histogram = helper::make_histogram($data);
+
+// Data for charts.
+// Each duration range forms its own line on the chart.
+
+$firstmonth = (int) userdate(strtotime("$monthstodisplay months ago"), '%Y%m'); // YYYYMM format.
```
keevan: Could we move the handling of this to a helper function? It seems to be used in multiple places, and would be good to avoid the manual cast and format setting.

e.g. `function get_year_month($timestamp): int` or similar.

Places this is used:
- page_group.php
- tests/tool_excimer_purge_page_group_test.php
- classes/task/purge_page_groups.php

page_group.php
```diff
@@ -0,0 +1,145 @@
...
+// Labels for the axis.
+$labels = [];
+
+// Label for each line.
+$linelabels = [];
+
+$month = $firstmonth;
+// We count through the months to get them in the right order, and to capture months that have no record in the database.
+for ($i = 0; $i < $monthstodisplay; ++$i) {
+    $labels[] = userdate(strtotime($month . '01'), '%b %Y');
```
keevan: Flagging this spot but similar to the previous suggestion for `get_month_formatted`

page_group.php
```diff
@@ -0,0 +1,145 @@
...
+        } else {
+            $durationseries[$rangehigh][] = 0;
+        }
+    }
+    // Increment the month, turning over the year if we get past December.
+    ++$month;
+    if ($month % 100 == 13) {
+        $month += 88;
+    }
+}
```
keevan: Could we move this into a function? It seems to be an isolated + complex enough and seems to only prepare `$durationseries` and `$linelabels`, making use of `$histograms` and the `$firstmonth` value.

page_group.php
```diff
@@ -0,0 +1,145 @@
...
+    $labels[] = userdate(strtotime($month . '01'), '%b %Y');
+    // For each histogram, the duration counts are stored in a simple array. We count from 0 to highest here to
+    // ensure that we have the same number of entries for each month.
+    for ($rangehigh = 0; $rangehigh < $highest; ++$rangehigh) {
+        if (isset($histograms[$month][$rangehigh])) {
+            $durationseries[$rangehigh][] = $histograms[$month][$rangehigh]['value'];
+            $linelabels[$rangehigh] = get_string('fuzzydurationrange', 'tool_excimer', $histograms[$month][$rangehigh]);
+        } else {
+            $durationseries[$rangehigh][] = 0;
+        }
```
keevan: Small suggestion. This change makes it so you're only pushing values into the array in 2 possible locations instead of 3, and removing the else statement.
```suggestion
        $rangehighvalue = 0;
        if (isset($histograms[$month][$rangehigh])) {
            $rangehighvalue = $histograms[$month][$rangehigh]['value'];
            $linelabels[$rangehigh] = get_string('fuzzydurationrange', 'tool_excimer', $histograms[$month][$rangehigh]);
        }
        $durationseries[$rangehigh][] = $rangehighvalue;
```

page_group.php
```diff
@@ -0,0 +1,145 @@
...
+
+$monthstodisplayurl = clone $url;
+$monthstodisplayurl->params(['id' => $pagegroupid]);
+$monthstodisplayselect = new \single_select(
+    $monthstodisplayurl,
+    'monthstodisplay',
+    array_slice(range(0, $expiry), 1, null, true),
+    $monthstodisplay
+);
+$monthstodisplayselect->set_label(get_string('months_to_display', 'tool_excimer'));
```
keevan: A comment for this block could be useful here. Took me a while to figure out it was a combo-box with a list of months (back) to display.

templates/page_group.mustache
jaypha: Which way would you prefer?

jaypha: I made it all have spacing.

templates/page_group.mustache
jaypha: done

classes/helper.php
jaypha: done. There are loose checks elsewhere that could be fixed as a separate PR.

page_group.php
jaypha: done

classes/helper.php
jaypha: done.

page_group.php
jaypha: The format itself is now part of the string table, and the whole thing moved to a helper function.

page_group.php
jaypha: A class 'monthint' was created to handle logic relating to this value. Feel free to suggest a better name.

page_group.php
jaypha: I'm not convinced that there is any benefit to doing so. I added some extra comments and simplified the code a bit to help the reader.

page_group.php
jaypha: done

page_group.php
jaypha: done

jaypha:
I forgot. Screenies!
Screenshot from 2022-09-28 12-06-36
Screenshot from 2022-09-28 12-36-35

page_group.php
keevan: I hope to look into it a bit more tomorrow, but quickly going over it, do you think it might be better served as a proper OO class? e.g. so you can do things like the below which might be useful come the pagination task:
- `$month->prev()->prev()->formatted()`
- `$month->prev()->prev()->timestamp()`
- `$month->prev()->prev()->month()`

I've omited the typical `get_` prefix but you get the point :)

PR #295, opened 2022-10-11
Fix missing cache definition string

cameron1729:


lang/en/tool_excimer.php
```diff
@@ -196,6 +196,7 @@
 
 // Miscellaneous.
 $string['cachedef_request_metadata'] = 'Excimer request metadata cache';
+$string['cachedef_page_group_metadata'] = 'Excimer page group metadata cache';
```
keevan, line 199: Thanks for making this change!

Can we also please bump the version.php :pray:

Issue #296, opened 2023-01-06
tool_excimer_after_config exception


aspark21:
Moodle 3.11 latest
Excimer latest

Exception calling 'tool_excimer_after_config'
line 49 of /admin/tool/excimer/classes/web_processor.php: call to tool_excimer\manager->get_starttime()
line 131 of /admin/tool/excimer/classes/manager.php: call to tool_excimer\web_processor->init()
line 51 of /admin/tool/excimer/lib.php: call to tool_excimer\manager->start_processor()
line 1064 of /lib/setup.php: call to tool_excimer_after_config()
line 192 of /config.php: call to require_once()
line 6 of /admin/upgradesettings.php: call to require_once()

jaypha:
The fix for #262 broke all versions before 4.1.

jaypha:
Because the call to tool_excimer_before_session_start() isn't being done, the manager is not setup properly, so the crash occurs.

brendanheywood:
@jaypha / @petersistrom can you please gracefully fix this in code, this came up in code review but fell through the cracks:

https://github.com/catalyst/moodle-tool_excimer/pull/262/files#r967772505

If before_session_start doesn't exist it should gracefully behave exactly as it did before

brendanheywood:
Fixed

PR #297, opened 2023-01-09
Issue #296: Fix missed call to manager::init().

jaypha:
https://github.com/catalyst/moodle-tool_excimer/issues/296

### DIFF
--- a/classes/manager.php
+++ b/classes/manager.php
@@ -80,12 +80,17 @@ class manager {
     /**
      * Get manager instance.
      *
+     * @param bool $initialise If true, will initialise the manager upon creation.
+     *
      * @return manager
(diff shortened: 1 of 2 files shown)


classes/manager.php
```diff
@@ -80,12 +80,17 @@ public function get_starttime(): float {
...
      * @return manager
      * @throws \dml_exception
      */
-    public static function get_instance(): manager {
+    public static function get_instance(bool $initialise = true): manager {
         if (!self::$instance) {
             self::create();
+            if ($initialise) {
+                self::$instance->init();
+            }
```
keevan, line 93: Nothing appears to call get_instance with $initialise = false, could we just always init it then?

PR #300, opened 2023-02-14
Issue #299: Fix bug causing insert ID error

### LINKED ISSUE #299: Race condition with interrupts diring a dml insert cause database write errors
Alex found a bug where we think the following is happening:

1) Some normal moodle code is doing an insert
```
$this->query_start($sql, $params, SQL_QUERY_INSERT);
```
2) the excimer time triggers an interrupt, which does its own internal logic and then tries to persist data using the moodle db api which is itself an insert which changes the state of the db handles attribute:

```
$id = @$this->mysqli->insert_id;
```

3) Then when it tries to continue doing the normal insert it started doing then the internal state has changed and it barfs.
https://github.com/moodle/moodle/blob/master/lib/dml/mysqli_native_moodle_database.php#L1452-L1454

The problem here is pretty well exactly the same as detecting we are inside a transaction which could be rolled back and the handling should be the same:

https://github.com/catalyst/moodle-tool_excimer/blob/8daedb09ee86dffec465f2021a891ab57dab2c55/classes/profile.php#L200-L214

I was originally going to say lets look at the current all stack to see if we are inside a dml function like insert_record_raw but perhaps the much simpler nuclear option is to do zero logic and *always* open a second fresh db connection. There are downsides to this, especially if the db connection takes a bit of time.

jaypha:
Resolves #299 

Remove all access to the database through the primary DB variable. Preload config values before processing.
Replace use of config caching with cache class.
Use alternative connection to access database during profile processing.

### DIFF
--- a/classes/profile_helper.php
+++ b/classes/profile_helper.php
@@ -27,6 +27,26 @@ namespace tool_excimer;
  */
 class profile_helper {
 
+    /** @var string Name used for group of all */
+    public const ALL_GROUP_CACHE_KEY = '__all__';
+
+    /** @var array Limits for storing profiles of a group. */
+    protected static $groupquotas;
+    /** @var array Limits for storing profiles. */
+    protected static $quotas;
+
+    /**
+     * Preload config values to avoid DB access during processing. See manager::get_altconnection() for more information.
+     */
+    public static function init() {
+        self::$groupquotas = [];
+        self::$quotas = [];
+        foreach (profile::REASON_STR_MAP as $reason => $reasonstr) {
+            self::$groupquotas[$reason] = (int) get_config('tool_excimer', 'num_' . $reasonstr . '_by_page');
+            self::$quotas[$reason] = (int) get_config('tool_excimer', "num_$reasonstr");
+        }
+    }
+
     /**
      * The number of profiles stored on disk.
      *
@@ -52,10 +72,7 @@ class profile_helper {
      * @return float
      */
     public static function get_min_duration_for_group_and_reason(string $group, int $reason, bool $usecache = true): float {
-        global $DB;
-
-        $reasonstr = profile::REASON_STR_MAP[$reason];
-        $pagequota = (int) get_config('tool_excimer', 'num_' . $reasonstr . '_by_page');
+        $pagequota = self::$groupquotas[$reason];
 
         // Grab the fastest profile for this page/request, and use that as
         // the lower boundary for any new profiles of this page/request.
@@ -69,14 +86,19 @@ class profile_helper {
             // the fact valid profiles will be added and the limits will be
             // breached for 'some time'. This will keep the constraints as
             // correct as possible.
-            $reasons = $DB->sql_bitand('reason', $reason);
+            $db = manager::get_altconnection();
+            if ($db === false) {
+                debugging('tool_excimer: Alt DB connection failed.');
+                return 0;
+            }
+            $reasons = $db->sql_bitand('reason', $reason);
             $sql = "SELECT duration as min_duration
                       FROM {tool_excimer_profiles}
                      WHERE $reasons != ?
                            AND groupby = ?
                   ORDER BY duration DESC
                      ";
-            $resultset = $DB->get_records_sql($sql, [
+            $resultset = $db->get_records_sql($sql, [
                 profile::REASON_NONE,
                 $group,
             ], $pagequota - 1, 1); // Will fetch the Nth item based on the quota.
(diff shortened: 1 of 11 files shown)


classes/profile_helper.php
```diff
@@ -104,51 +119,34 @@ public static function get_min_duration_for_group_and_reason(string $group, int
...
+        $cachekey = '__all__';
+        $cachefield = 'profile_type_' . $reason . '_min_duration_s';
+        $cache = \cache::make('tool_excimer', 'request_metadata');
+        $result = $cache->get($cachekey);
 
-        if (!$usecache || $result === false) {
+        if (!$usecache || $result === false || !isset($result[$cachefield])) {
             // Get and set cache.
-            $reasons = $DB->sql_bitand('reason', $reason);
+            $db = manager::get_altconnection();
```
brendanheywood, line 137: do we need protection here if no connection like here:

https://github.com/catalyst/moodle-tool_excimer/pull/300/files#diff-6546d87581281c290f904eb79396be0da45288aec26b4159c0a102d85118872fR189

classes/profile_helper.php
```diff
@@ -104,51 +119,34 @@ public static function get_min_duration_for_group_and_reason(string $group, int
...
-    }
-
-    /**
-     * Clears the plugin cache for keys used for the provided reasons
-     *
-     * @author Kevin Pham <kevinpham@catalyst-au.net>
-     *
-     * @param int $reason bitmap of reason(s)
-     */
-    public static function clear_min_duration_cache_for_reason(int $reason): void {
```
brendanheywood, line 144: is it worth keeping this method but just changing its implementation?

classes/script_metadata.php
```diff
@@ -83,6 +83,32 @@ class script_metadata {
...
+    protected static $stacklimit;
+    /** @var int Sample limit config. */
+    protected static $samplelimit;
+    /** @var int Sampling period */
+    public static $samplems;
+    /** @var string */
+    protected static $redactparams;
+
+    /**
+     * Preload config values to avoid DB access during processing.
```
brendanheywood: maybe add another sentence around why this is so important

tests/tool_excimer_script_metadata_test.php
```diff
@@ -193,6 +198,7 @@ public function sampling_limit_provider(): array {
      */
     public function test_get_redactable_param_names(string $config, array $expected) {
         set_config('redact_params', $config, 'tool_excimer');
+        script_metadata::init();
         $this->assertEquals($expected, script_metadata::get_redactable_param_names());
```
brendanheywood, line 202: is it viable in any way to add a unit test around the insert inside an insert or it that too hairy?

classes/profile_helper.php
jaypha: The delete statements that replace it are all one-liners.

I noticed that the way the cache was implemented is sub-optimal, and want to create a separate issue for it. For the purpose of this bugfix, however, this is sufficient.

classes/profile_helper.php
jaypha: I will fix this.

classes/script_metadata.php
jaypha: There is already some documentation on this, but I will expand on it.

tests/tool_excimer_script_metadata_test.php
jaypha: A problem with testing this is that the database does not get reset after each unit test when using the alternate connection. If this is important to include, then I can spend some time researching how to do this.

classes/profile_helper.php
```diff
@@ -104,51 +123,38 @@ public static function get_min_duration_for_group_and_reason(string $group, int
...
     public static function get_min_duration_for_reason(int $reason, bool $usecache = true): float {
-        global $DB;
-
-        $reasonstr = profile::REASON_STR_MAP[$reason];
-        $quota = (int) get_config('tool_excimer', "num_$reasonstr");
+        $quota = self::$quotas[$reason];
 
-        $cachekey = 'profile_type_' . $reason . '_min_duration_s';
-        $result = get_config('tool_excimer', $cachekey);
+        $cachekey = '__all__';
```
keevan: Would this be better if we change this to a class constant, and update all relevant references to match?

classes/profile_helper.php
brendanheywood: ok cool

tests/tool_excimer_script_metadata_test.php
brendanheywood: No I was expecting it to be a weird thing to try and unit test

keevan (approved):
Looks good to me.

PR #303, opened 2023-02-16
Update xmldb to fix Totara unit test errors

petersistrom:


keevan (approved):
LGTM.

Thanks for fixing that up Peter

PR #309, opened 2023-04-28
Fixes for php 8.1 compatibility

SashaAnastasi:


SashaAnastasi:
This PR addresses the following bugs:

++ **PHP Deprecated:** strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated
_classes/manager.php on line 264,265,266_


++ **PHP Deprecated:**  Automatic conversion of false to array is deprecated
_classes/profile_helper.php on line 109,155_


++ **TypeError:** tool_excimer\script_metadata::get_sample_limit(): Return value must be of type int, null returned
++ Trying to access array offset on value of type null
_tests/tool_excimer_profile_test.php_

classes/profile_helper.php
```diff
@@ -79,7 +79,7 @@ public static function get_min_duration_for_group_and_reason(string $group, int
         $cachekey = $group;
         $cachefield = 'min_duration_for_reason_' . $reason . '_s';
         $cache = \cache::make('tool_excimer', 'request_metadata');
-        $result = $cache->get($cachekey);
+        $result = $cache->get($cachekey) ?: array();
 
         if (!$usecache || $result === false || !isset($result[$cachefield])) {
```
golenkovm: As `$result` will never be `false` shouldn't we update `$result === false` too?

classes/profile_helper.php
```diff
@@ -130,7 +130,7 @@ public static function get_min_duration_for_reason(int $reason, bool $usecache =
 
         $cachefield = 'profile_type_' . $reason . '_min_duration_s';
         $cache = \cache::make('tool_excimer', 'request_metadata');
-        $result = $cache->get(self::ALL_GROUP_CACHE_KEY);
+        $result = $cache->get(self::ALL_GROUP_CACHE_KEY) ?: array();
 
         if (!$usecache || $result === false || !isset($result[$cachefield])) {
```
golenkovm: Same suggestion here

golenkovm:
Hi @SashaAnastasi 

Thank you for submitting this pull request. I left a couple of minor notes. Could you please address them once you have time? 

As I can see the third error (TypeError, Return value must be of type int, null returned) has been already fixed as part of https://github.com/catalyst/moodle-tool_excimer/commit/d0ee314e8a53d1bf2278f0f5a793671c709b7b24 which is good.

Also, may I ask you to tweak CI matrix and add php8.1 so we could make sure it's all passing with php8.1?

PS: It's worth to mention that the patch is fully compatible with previous old moodle/php version so we will be able to merge this to `MOODLE_35_STABLE` branch :+1: 

Cheers,
Misha

SashaAnastasi:
Thanks for that @golenkovm , I've traded out the condition to test for falsiness instead.
Could you please rereview?

-Sasha

golenkovm:
Thanks @SashaAnastasi 
The patch is looking good to me.
Speaking of the CI change and php8.1 it seems like this should be done as part of a separate pull request in https://github.com/catalyst/catalyst-moodle-workflows so I'm happy to merge this now.

Cheers,
Misha

PR #313, opened 2023-05-02
310 log what course a profile belongs to

### LINKED ISSUE #310: Log what course a profile belongs to
Often you can see a bunch of different pages which are slow and there isn't an obvious pattern, but when you dig into it you notice that all of the profiles are in the same course but might be for different profiles. 

We should make this simpler to see:
* [ ]  log the course for a profile. Note the $COURSE might not be set until after a few ticks (like other things we collect) so we need to check this on each tick if it isn't set yet
* [ ] show this in the drill down page as a new column eg /admin/tool/excimer/slowest_web.php?group=mod%2Fquiz%2Freport.php
* [ ] make a new top level excimer tab which just shows all aggregates profiles the same as /admin/tool/excimer/slowest_web.php, but instead of aggregated by profile it is grouped by the course
* [ ] have a drill down page for each course. Both this and the page above could be new pages, or the existing pages with more params

### LINKED ISSUE #312: Tests executed individually do not correctly setup data
E.g. if you run `vendor/bin/phpunit --testsuite=tool_excimer_testsuite --filter=test_reasons_are_being_stored` - you get an error 

```
There was 1 error:

1) tool_excimer\tool_excimer_profile_test::test_reasons_are_being_stored
Trying to access array offset on value of type null

/var/www/evidence-central/admin/tool/excimer/classes/profile_helper.php:75
/var/www/evidence-central/admin/tool/excimer/classes/profile.php:238
/var/www/evidence-central/admin/tool/excimer/tests/tool_excimer_profile_test.php:89
/var/www/evidence-central/admin/tool/excimer/tests/tool_excimer_profile_test.php:241
/var/www/evidence-central/lib/phpunit/classes/advanced_testcase.php:80
phpvfscomposer:///var/www/evidence-central/vendor/phpunit/phpunit/phpunit:97
```

This is because `profile_helper::init` has not been called yet, and its trying to access a static array https://github.com/catalyst/moodle-tool_excimer/blob/619199f7b82141387c2683910ee88df486780adf/classes/profile_helper.php#L34. The fix for this is fairly trivial, just need to call: 

https://github.com/catalyst/moodle-tool_excimer/blob/619199f7b82141387c2683910ee88df486780adf/classes/profile_helper.php#L41-L48

Before the tests.

matthewhilton:
Closes #310 
Closes #312 

**Changes**

- Bugfix for running unit tests individually ( #312 )

- Logs $COURSE->id when saving profile
- New tab for group by courses (similar to the group by script)
- Drill down page for courses
- Individual profile view shows course

**Screenshots**
Course groups
image

Course drill-down
image

Individual profile
image


classes/helper.php
```diff
@@ -17,6 +17,7 @@
 namespace tool_excimer;
 
 use core_filetypes;
+use moodle_url;
```
Peterburnett: nitpick: I generally prefer to just \ this inline, purely because its a little cleaner IMO.

classes/output/tabs.php
```diff
@@ -60,6 +60,12 @@ public function export_for_template(\renderer_base $output): array {
                     'title' => get_string('report_slowest_other', 'tool_excimer'),
                     'text' => get_string('tab_slowest_other', 'tool_excimer'),
                 ],
+                [
+                    'id' => 'course',
+                    'link' => [['link' => new \moodle_url('/admin/tool/excimer/slow_course.php')]],
```
Peterburnett, line 65: Why is the link a double array? Im looking at the wider pattern and it seems really bizarre

classes/profile.php
```diff
@@ -212,6 +212,10 @@ public function save_record(): int {
             $this->raw_set('userid', $USER->id);
         }
 
+        if ($this->check_update_courseid($COURSE->id)) {
```
Peterburnett, line 215: Is this going to store everything in the admin level under course=1? That might be fine, or it might muddy the waters a little

classes/slow_course_table.php
```diff
@@ -0,0 +1,162 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_excimer;
+
+/**
+ * Table for displaying grouped profile lists.
+ *
+ * @package   tool_excimer
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
```
Peterburnett: Author needs adjusting

Peterburnett: Also year

classes/slow_course_table.php
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     * Sets the SQL for the table.
+     */
+    protected function put_sql(): void {
+        list($filterstring, $filterparams) = $this->get_filter_for_sql();
+
+        $this->set_count_sql(
+            "SELECT count(distinct request)
```
Peterburnett: Just for readability, all the SQL keywords should be caps eg distinct -> DISTINCT as well as the functions in the next query (MAX, COUNT) etc

lang/en/tool_excimer.php
```diff
@@ -142,6 +144,7 @@
 $string['field_hostname'] = 'Host name';
 $string['field_useragent'] = 'User agent';
 $string['field_versionhash'] = 'Version Hash';
+$string['field_course'] = 'Course';
```
Peterburnett: This and 'Courses' I expect would already exist in core, and can be reused

slow_course.php
```diff
@@ -0,0 +1,52 @@
...
+ * @author    Matthew Hilton <matthewhilton@catalyst-au.net>
+ * @copyright 2023, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+use tool_excimer\profile_table;
+use tool_excimer\profile_table_page;
+use tool_excimer\slow_course_table;
+
+require_once('../../../config.php');
```
Peterburnett: Please make this absolute with `__DIR__`, i've seen this have issues before. PHP sometimes screws this up

tests/tool_excimer_helper_test.php
```diff
@@ -0,0 +1,61 @@
...
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ * @covers \tool_excimer\helper
+ */
+class tool_excimer_helper_test extends \advanced_testcase {
+    /**
+     * Set up before each test
+     */
+    protected function setUp(): void {
+        parent::setUp();
+        $this->resetAfterTest();
```
Peterburnett: Can we just put this in the test itself? that way additional tests aren't forced to use it if its not required

tests/tool_excimer_profile_test.php
```diff
@@ -33,6 +33,8 @@ class tool_excimer_profile_test extends \advanced_testcase {
     protected function setUp(): void {
         parent::setUp();
         $this->resetAfterTest();
+        profile_helper::init();
```
Peterburnett, line 36: Why are these required. It doesnt look like you are doing anything new in your tests

classes/slow_course_table.php
```diff
@@ -0,0 +1,162 @@
...
+        return \html_writer::link($url, helper::course_display_name($record->courseid));
+    }
+
+    /**
+     * Max created column.
+     *
+     * @param \stdClass $record
+     * @return string
+     */
+    public function col_maxcreated(\stdClass $record): string {
```
brendanheywood: can we inherit a lot of these columns from another table class and some some boilerplate?

classes/output/tabs.php
matthewhilton: Not sure - it seems redundant, since at the bottom it just calls `link[0]->link`

https://github.com/catalyst/moodle-tool_excimer/blob/5257c205ba836337b01b9f5f2780efea73332bcc/classes/output/tabs.php#L90-L94

matthewhilton: Was added in 5a1bad9 - maybe the idea was multiple links per tab?

classes/profile.php
matthewhilton: Yes correct. I would expect anyone using the Excimer plugin has enough Moodle knowledge to know why course id = 1 is used so often so I don't think it's an issue.

lang/en/tool_excimer.php
matthewhilton: We can't reuse them since the excimer plugin looks for `get_string('field_' . $fieldname)`

matthewhilton: https://github.com/catalyst/moodle-tool_excimer/blob/5257c205ba836337b01b9f5f2780efea73332bcc/classes/profile_table.php#L90

tests/tool_excimer_profile_test.php
matthewhilton: Its for #312 - Because otherwise the tests fail because the static vars are not populated yet https://github.com/catalyst/moodle-tool_excimer/blob/5257c205ba836337b01b9f5f2780efea73332bcc/classes/profile_helper.php#L34-L36

I split it into a separate commit 83c3311d0509c562ec3793f73d3b2aef64f5098d but kept it in this Pr since its a tiny change

classes/slow_course_table.php
matthewhilton: Ah probably. I did reuse a lot of code from `grouped_profile_table.php` - inheriting it is probably cleaner. Will update...

matthewhilton:
I have adjusted the code a bit. Now it is structured:

- `grouped_profile_table` is subclasses by `grouped_course_profile_table` and `grouped_script_profile_table`. Each just specifies what column is being grouped and the parent class does the rest.
- Renamed the db column `groupby` -> `scriptgroup` since it was a bit confusing now that we group by more than just the script. Updated all the references to groupby

brendanheywood:
Functionally I think this is good to go. UX wise I think there should be an extra drill down between the course tab and the page tab, eg:

1) I start here: https://master.localhost/admin/tool/excimer/slow_course.php
image

2) Then I drill down into a course and see a list of individual profiles:
https://master.localhost/admin/tool/excimer/slow_course.php?courseid=4
image

As you can see there is lots of duplicates. I think it would be better if instead it showed the same contents as slowest_web.php but filtered to that course, and then you drill down again to the profiles. 

Maybe this is better as slowest_web.php?courseid=4 or maybes its better inside slow_course.php I don't mind

matthewhilton:
Nice idea @brendanheywood - very valid. Probably would need to restructure the pages a bit to accomplish that so probably best for a followup issue.

PR #318, opened 2023-06-29
docs: Add manual profiling section to docs

matthewhilton:
- Updated docs regarding how to manually trigger profiling
- My local markdown linter for VSCode automatically updated the table of contents - let me know if I should revert this but I think it would be good to have it in sync


keevan (approved):
Changes here look good. I am also concerned about keeping the TOC in sync since not everyone uses VSCode. 

Ideally, the readme should be succinct so the TOC isn't required, but it's nice to have.

It'd be nice to fix up the ci badge as well whilst you're here, but otherwise I'm happy for this to be merged.

keevan:
Approving changes. The TOC out of sync concern is a small concern, and the new information is important than the table coming out of sync every now and then.

PR #327, opened 2023-08-29
[#326] change field referer to text

### LINKED ISSUE #326: Referrer URL is too long for the field
Just came across this error 

```
Error writing to database 
...
Data too long for column 'referer' at row 1
INSERT INTO mdl_tool_excimer_profiles...
```

Looks like if the referring url is greater than 256 characters, which is possible as max url length for various browsers is much bigger than 256 characters, it will cause a fatal error.

We should probably make this field as TEXT rather than limited to 256 characters?
https://github.com/catalyst/moodle-tool_excimer/blob/MOODLE_35_STABLE/db/install.xml#L26

jwalits:
Fixes #326


keevan (approved):
Looks good, I know it's unrelated, but are you able to take a look at the failing CI as well? It seems like the duplicate code check is complaining, and it might be a quick one to resolve.

jwalits:
> Looks good, I know it's unrelated, but are you able to take a look at the failing CI as well? It seems like the duplicate code check is complaining, and it might be a quick one to resolve.

Sure, I'll try and have a look at those today......thanks for reviewing @keevan

PR #328, opened 2023-08-31
[#272] Fix for fuzzy grouping, function rename

Fragonite:
Assumes paths should not have a trailing slash (unless the path is empty). I can change this behaviour if required.

References: 
Issue #272
PR #285

### DIFF
--- a/classes/script_metadata.php
+++ b/classes/script_metadata.php
@@ -205,23 +205,40 @@ class script_metadata {
     }
 
     /**
-     * Gets the name of the script.
-     *
+     * Gets the name of the script. Parameters share the names of globals as a hint to the caller.
+     * @param string|null $me
+     * @param string|null $script
      * @return string the request path for this profile.
      */
-    public static function get_request(): string {
-        global $SCRIPT, $ME, $CFG;
-
-        if (!isset($ME)) {
-            // If set, it will trim off the leading '/' to normalise web & cli requests.
-            $request = isset($SCRIPT) ? ltrim($SCRIPT, '/') : self::REQUEST_UNKNOWN;
-            return $request;
+    public static function get_normalised_relative_script_path($me, $script): string {
+        if (isset($me)) {
+            $scriptpath = $me;
+        } else if (isset($script)) {
+            $scriptpath = $script;
+        } else {
+            return self::REQUEST_UNKNOWN;
         }
 
-        $request = (new \moodle_url($ME))->out_omit_querystring();
-        $request = str_replace($CFG->wwwroot, '', $request);
-        $request = ltrim($request, '/');
-        return $request;
+        // Remove consecutive slashes.
+        $scriptpath = preg_replace('/\/+/', '/', $scriptpath);
+
+        // Strip pathinfo.
+        $scriptpath = preg_replace('/\.php.*$/', '.php', $scriptpath, 1);
+
+        // Strip off the query string ($ME includes this).
+        $scriptpath = preg_replace('/\?.*$/', '', $scriptpath);
+
+        // Remove 'index.php' from the end ($SCRIPT includes this).
+        $scriptpath = preg_replace('/\/index\.php$/', '', $scriptpath, 1);
+
+        // Remove leading and trailing slashes.
+        $scriptpath = trim($scriptpath, '/');
+
+        if ($scriptpath === '') {
+            return '/';
+        }
+
+        return $scriptpath;
     }
 
     /**
@@ -401,7 +418,7 @@ class script_metadata {
     public static function get_sampling_period(): float {
         $period = get_config('tool_excimer', 'sample_ms') / 1000;
         $insensiblerange = $period >= self::SAMPLING_PERIOD_MIN && $period <= self::SAMPLING_PERIOD_MAX;
-        if (! $insensiblerange) {
+        if (!$insensiblerange) {
             set_config('sample_ms', self::SAMPLING_PERIOD_DEFAULT * 1000, 'tool_excimer');
             $period = self::SAMPLING_PERIOD_DEFAULT;
         }
--- a/tests/tool_excimer_script_metadata_test.php
+++ b/tests/tool_excimer_script_metadata_test.php
@@ -189,28 +189,6 @@ class tool_excimer_script_metadata_test extends \advanced_testcase {
     }
 
     /**
-     * Tests script_metadata::get_request().
-     * @dataProvider relative_script_path_provider
-     * @covers \tool_excimer\script_metadata::get_request
-     * @param string $path
-     * @param string $expected
-     */
-    public function test_get_request(string $path, string $expected) {
-        script_metadata::init();
-
-        global $SCRIPT, $ME;
-        $globalscript = $SCRIPT ?? null;
-        $globalme = $ME ?? null;
-
-        $SCRIPT = $path;
-        $ME = $path;
-        $this->assertEquals($expected, script_metadata::get_request());
-
-        $SCRIPT = $globalscript;
-        $ME = $globalme;
-    }
-
-    /**
      * Tests script_metadata::get_normalised_relative_script_path().
      * @dataProvider relative_script_path_provider
      * @covers \tool_excimer\script_metadata::get_normalised_relative_script_path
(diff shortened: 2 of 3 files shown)


classes/script_metadata.php
```diff
@@ -209,18 +209,38 @@ public static function get_redactable_param_names(): array {
      *
      * @return string the request path for this profile.
      */
-    public static function get_request(): string {
-        global $SCRIPT, $ME, $CFG;
+    public static function get_normalised_relative_script_path(): string {
```
brendanheywood: is there a reason to rename it?

classes/script_metadata.php
```diff
@@ -209,18 +209,38 @@ public static function get_redactable_param_names(): array {
...
+        $request = $SCRIPT;
+        $count = 0;
+
+        // Remove multiple consecutive slashes.
+        do {
+            $request = str_replace('//', '/', $request, $count);
+        } while ($count);
+
         $request = ltrim($request, '/');
+        $request = rtrim($request, '/');
```
brendanheywood: these two lines could be combined into just 'trim'

classes/script_metadata.php
```diff
@@ -209,18 +209,38 @@ public static function get_redactable_param_names(): array {
...
-        global $SCRIPT, $ME, $CFG;
+    public static function get_normalised_relative_script_path(): string {
+        global $SCRIPT;
 
-        if (!isset($ME)) {
-            // If set, it will trim off the leading '/' to normalise web & cli requests.
-            $request = isset($SCRIPT) ? ltrim($SCRIPT, '/') : self::REQUEST_UNKNOWN;
-            return $request;
+        // This is from the previous iteration of the function.
+        if (!isset($SCRIPT)) {
```
brendanheywood: cam you please cover this with unit tests, the removal of the logic around $ME seems a bit risky to me, it might be fine but having the tests proves it and I think there is some edge cases here which we don't want to regress

This would be a good candidate for a dataProvider, see the existing tests for an example

classes/script_metadata.php
```diff
@@ -209,18 +209,38 @@ public static function get_redactable_param_names(): array {
...
         }
 
-        $request = (new \moodle_url($ME))->out_omit_querystring();
-        $request = str_replace($CFG->wwwroot, '', $request);
+        $request = $SCRIPT;
+        $count = 0;
+
+        // Remove multiple consecutive slashes.
+        do {
+            $request = str_replace('//', '/', $request, $count);
```
brendanheywood: maybe some personal preference here but I reckon this might be simpler as a preg_replace without needing a loop

classes/script_metadata.php
```diff
@@ -209,18 +209,38 @@ public static function get_redactable_param_names(): array {
...
+        } while ($count);
+
         $request = ltrim($request, '/');
+        $request = rtrim($request, '/');
+
+        $indexphp = 'index.php';
+
+        // Trim "index.php" from the end of the request. Case sensitive.
+        if (substr($request, -(strlen($indexphp))) === $indexphp) {
```
brendanheywood: maybe another candidate to be simpler by using a regex

Fragonite: We can make it part of the same preg_replace as above.

classes/script_metadata.php
Fragonite: I thought about it and I agree.

classes/script_metadata.php
Fragonite: Definitely needs unit tests but how would they realistically cover the interaction between $SCRIPT and $ME?

classes/script_metadata.php
Fragonite: #285 renames it. I do think the name should be more descriptive and I am open to suggestions.

brendanheywood: ok cool new name is good

classes/script_metadata.php
brendanheywood: Currently get_normalised_relative_script_path is not a pure function as it has an input the value inside $SCRIPT, and it did previously also use $ME. I think that there is some weird edge cases where the value of $ME should be leveraged and this should have been covered by unit tests before so we should fix that and make sure they are covered. Off the top of my head one edge case to test for is when you go to a page which doesn't exist, like /does-not-exist then nginx served this as an error page which is /error/index.php and I think it needs both globals to correctly handle that.

As for the unit tests, there is two approaches and both would have the value of $SCRIPT as the first input, the value of $ME as the second, and the third would be the expected outcome
1) in the unit test you simply set $SCRIPT and $ME to what the data provider give it and then run the function. You should probably stash the original values and reset them so other later unit tests don't get weird side effects.
2) we remove the global dependency on $SCRIPT and $ME and make these explicit function parameters eg $request = script_metadata::get_normalised_relative_script_path($SCRIPT, $ME); - this one is the more proper approach but you'll find lots of examples of the first approach in various core unit tests

tests/tool_excimer_script_metadata_test.php
```diff
@@ -188,6 +188,48 @@ public function sampling_limit_provider(): array {
...
+    public function test_get_normalised_relative_script_path(string $pathinfo, string $expected) {
+        script_metadata::init();
+        $this->assertEquals($expected, script_metadata::get_normalised_relative_script_path($pathinfo));
+    }
+
+    /**
+     * Provider for test_get_normalised_relative_script_path().
+     * @return \string[][]
+     */
+    public function normalised_relative_script_path_provider(): array {
```
brendanheywood: this is a great start :)

it's probably worth adding this unit test as a first logic commit so you can run it on the old function and then compare to the new function to ensure that nothing has changed unintentionally after you redo the logic of get_normalised_relative_script_path

There are a couple things which I'm pretty sure have changed and should not, like the original function did an ltrim on /, so '/home' should result in 'home' and the only time we want '/' at the start was if the string was empty. So we should explicitly see those changed expectations in the second commits dataprovider changes

classes/script_metadata.php
```diff
@@ -206,22 +206,35 @@ public static function get_redactable_param_names(): array {
...
+            // This else block normalises $scriptpath to match the format of $SCRIPT.
+
+            $scriptpath = '/' . $scriptpath;
+            $scriptpath = preg_replace('/\/+/', '/', $scriptpath);
+            $scriptpath = preg_replace('/\/$/', '/index.php', $scriptpath, 1);
+        }
+
+        $scriptpath = preg_replace('/\/index\.php$/', '', $scriptpath, 1);
+
+        if ($scriptpath === '') {
```
brendanheywood, line 237: also should get_groupby_value() be touched?

Fragonite: The name could use another underscore and there is a TODO in `SCRIPT_NAMES_FOR_GROUP_REFINING` but otherwise it looks fine. Did you want me to change the name?

tests/tool_excimer_script_metadata_test.php
Fragonite: The leading slash was new logic to make it more consistent with $SCRIPT and the empty case but we can definitely change it back.

I will add the old function back in and compare unit tests.

classes/script_metadata.php
Fragonite: It is now a pure function with some added heuristics for better normalisation.

Fragonite:
Checks are passing after removing type hinting from `get_normalised_relative_script_path()`. Checks were passing locally, perhaps the CI code needs updating?

brendanheywood:
@Fragonite no the CI is correct, it looks at the version.php file to determine which moodle stables it should be run on, and the rules depend on the stables, see this line:

https://github.com/catalyst/moodle-tool_excimer/blob/MOODLE_35_STABLE/version.php#L31

So because we support older stables we need to run the tests on each stable in whatever that stables wants in terms of php version and other versions like db version. If we wanted to use the latest and greatest php features then we would cut a new branch eg MOODLE_401_STABLE and make the old one finish support at 4.0 and the new one takes over. We generally try to support as much as possible on each branch until we are forced to branch

classes/script_metadata.php
```diff
@@ -224,6 +224,44 @@ public static function get_request(): string {
...
     }
 
+    /**
+     * Gets the name of the script.
+     * Parameters are share the names of globals as a hint to the caller.
+     * @param ?string $ME
+     * @param ?string $SCRIPT
+     * @return string the request path for this profile.
+     */
+    public static function get_normalised_relative_script_path(?string $ME, ?string $SCRIPT): string {
```
brendanheywood: this makes a new method but doens't remove the old one get_request which should now be unused code after this commit?

Can you please also squash this commit with the ones which come after it as they are all logically part of the same change and then we should be good to go

PR #331, opened 2023-12-04
Fix for unhandled dml_exception (get_min_duration_for_reason with missing table)

### LINKED ISSUE #330: dml_read_exception when reading from non-existent mdl_tool_excimer_profiles table
In some cases the `mdl_tool_excimer_profiles` table might not exist before `profile_helper.php` runs, which can result in the above exception.

This is happening when the database is being refreshed.

Fragonite:
Resolves #330

### DIFF
--- a/classes/web_processor.php
+++ b/classes/web_processor.php
@@ -74,17 +74,27 @@ class web_processor implements processor {
         $this->profile->set('created', $this->sampleset->starttime);
 
         $manager->get_timer()->setCallback(function () use ($manager) {
-            $this->process($manager, false);
+            try {
+                $this->process($manager, false);
+            } catch (\dml_exception $e) {
+                // Ignore errors during processing.
+                debugging('tool_excimer: Timer callback failed: ' . $e->getMessage());
+            }
         });
 
         \core_shutdown_manager::register_function(
             function () use ($manager) {
-                $manager->get_timer()->stop();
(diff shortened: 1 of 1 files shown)


classes/profile_helper.php
```diff
@@ -145,9 +145,19 @@ public static function get_min_duration_for_reason(int $reason, bool $usecache =
...
+            try {
+                $resultset = $db->get_records_sql(
+                    $sql,
+                    [profile::REASON_NONE],
+                    $quota - 1, // Will fetch the Nth item based on the quota.
+                    1
+                );
+            } catch (\dml_exception $e) {
+                debugging('tool_excimer: Failed to get min duration: '.$e->getMessage());
+                return null;
```
brendanheywood: the calling code expects a number and hasn't been touched. Thinking that we could just return a very large number like 1 week so everything is smaller than it

Fragonite: I checked the calling code and it all looks compatible. Returning a large number could also work. See my alternate fix which could be a better way of solving the problem.

Fragonite:
This issue can be closed. It turns out the problem that prompted the creation of this issue was related to Moodle MUC which was causing other problems too.

brendanheywood:
Closing as apparently not needed 

https://github.com/catalyst/moodle-tool_excimer/issues/330#issuecomment-1844067068

PR #339, opened 2024-04-17
Update probabilistic counting #336

### LINKED ISSUE #336: Probabilistic counting of pages is not working correctly

bwalkerl:
Closes #336 

The updated logic should cover the intent mentioned on the issue, though this will have slightly different results compared to the previous implementation.

### DIFF
--- a/classes/manager.php
+++ b/classes/manager.php
@@ -345,26 +345,31 @@ class manager {
 
     /**
      * Increments a counter using the approximate counting algorithm.
-     * Can handle up to PHP_INT_SIZE_BITS counts (2^PHP_INT_SIZE_BITS events).
      *
      * See https://en.wikipedia.org/wiki/Approximate_counting_algorithm
      *
      * @param int $current The current count.
+     * @param int $increment The raw increment in count.
+     * @param float|null $random random number between 0 and 1, for unit testing.
      * @return int The new count.
      */
-    public static function approximate_increment(int $current): int {
-        // If the number of events is ever expected to be more than 2 billion, a refactor may be needed.
+    public static function approximate_increment(int $current, int $increment = 1, ?float $random = null): int {
+        $approx = 2 ** $current;
+        // If we are incrementing by more than the current approximation, we can bump the count.
+        if ($increment > $approx) {
+            $total = $increment + $approx;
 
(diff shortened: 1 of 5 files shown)


classes/manager.php
```diff
@@ -345,26 +345,25 @@ public function is_considered_slow(profile $profile): bool {
...
-        $bits = $bits << (self::PHP_INT_SIZE_BITS - $current);
-
-        // If the bits are all zero (equiv to all coin tosses = tails), then we increment the counter.
-        if ($bits == 0) {
-            return $current + 1;
+    public static function approximate_increment(int $current, int $increment = 1): int {
+        $approx = 2 ** $current;
+        // If we are incrementing by more than the current approximation, we can bump the count.
+        if ($increment > $approx) {
+            return self::approximate_increment($current + 1, $increment - $approx);
```
brendanheywood: Can this be done without recursion? Let’s say current = 4 so approx 16, and we increment by 128, then we are going g to turn either 7 most of the time and sometimes 8. So we can either do that maths directly, or call increment(log2($increment), $current ** 2) which should never recurse a second time

bwalkerl:
Latest pushes include:
- Updating unit test to be deterministic
- Fixing a bug where the approximation was starting at 2 instead of 1. The display seemed to be intentionally using lower values because of this, so that has also been adjusted (including removing the range - because the way the range was calculated didn't make sense).

Issue #357, opened 2024-09-04
Upgrade error on latest


x-iy:
On mysql 8, php 7.4 get error on Moodle upgrade:

```
-->tool_excimer
!!! Error writing to database !!!
!! You can't specify target table 'mdl_tool_excimer_page_groups' for update in FROM clause
DELETE FROM mdl_tool_excimer_page_groups
                 WHERE id NOT IN (
                     SELECT MIN(id)
                       FROM mdl_tool_excimer_page_groups
                   GROUP BY LOWER(name), month
                 )
[array (
)]
Error code: dmlwriteexception !!
!! Stack trace: * line 497 of /lib/dml/moodle_database.php: dml_write_exception thrown
* line 293 of /lib/dml/moodle_read_slave_trait.php: call to moodle_database-&gt;query_end()
* line 1181 of /lib/dml/mysqli_native_moodle_database.php: call to mysqli_native_moodle_database-&gt;query_end()
* line 506 of /admin/tool/excimer/db/upgrade.php: call to mysqli_native_moodle_database-&gt;execute()
* line 789 of /lib/upgradelib.php: call to xmldb_tool_excimer_upgrade()
* line 1967 of /lib/upgradelib.php: call to upgrade_plugins()
* line 274 of /admin/cli/upgrade.php: call to upgrade_noncore()
 !!
```

x-iy:
Fixed with https://github.com/catalyst/moodle-tool_excimer/pull/358

PR #358, opened 2024-09-04
Fix for #357 - error on moodle upgrade mysql 8

x-iy:
Fix for #357


keevan (approved):
Thanks, this works in pgsql - tested locally.

PR #360, opened 2024-09-06
perf: adjust handling based on real data where single DELETE was bottleneck

keevan:
Replace 'DELETE FROM where id NOT IN' with a SELECT and DELETE statement where id is in a known scope, better utilising indexes

Here is the explain for the DELETE prior to this patch:
```
+----+--------------------+------------------------------+------------+------+---------------+------+---------+------+-------+----------+---------------------------------+
| id | select_type        | table                        | partitions | type | possible_keys | key  | key_len | ref  | rows  | filtered | Extra                           |
+----+--------------------+------------------------------+------------+------+---------------+------+---------+------+-------+----------+---------------------------------+
|  1 | DELETE             | mdl_tool_excimer_page_groups | NULL       | ALL  | NULL          | NULL | NULL    | NULL | 50341 |   100.00 | Using where                     |
|  2 | DEPENDENT SUBQUERY | <derived3>                   | NULL       | ALL  | NULL          | NULL | NULL    | NULL | 50341 |   100.00 | Using temporary; Using filesort |
|  3 | DERIVED            | mdl_tool_excimer_page_groups | NULL       | ALL  | NULL          | NULL | NULL    | NULL | 50341 |   100.00 | NULL                            |
+----+--------------------+------------------------------+------------+------+---------------+------+---------+------+-------+----------+---------------------------------+
```

Here's an example of the difference between having a NOT IN vs an IN check, with dummy data:
```
mysql> explain      DELETE FROM mdl_tool_excimer_page_groups              WHERE id in (1,2,3);
+----+-------------+------------------------------+------------+-------+---------------+---------+---------+-------+------+----------+-------------+
| id | select_type | table                        | partitions | type  | possible_keys | key     | key_len | ref   | rows | filtered | Extra       |
+----+-------------+------------------------------+------------+-------+---------------+---------+---------+-------+------+----------+-------------+
|  1 | DELETE      | mdl_tool_excimer_page_groups | NULL       | range | PRIMARY       | PRIMARY | 8       | const |    3 |   100.00 | Using where |
+----+-------------+------------------------------+------------+-------+---------------+---------+---------+-------+------+----------+-------------+
1 row in set (0.00 sec)

mysql> explain      DELETE FROM mdl_tool_excimer_page_groups              WHERE id NOT in (1,2,3);
+----+-------------+------------------------------+------------+-------+---------------+---------+---------+-------+-------+----------+-------------+
| id | select_type | table                        | partitions | type  | possible_keys | key     | key_len | ref   | rows  | filtered | Extra       |
+----+-------------+------------------------------+------------+-------+---------------+---------+---------+-------+-------+----------+-------------+
|  1 | DELETE      | mdl_tool_excimer_page_groups | NULL       | range | PRIMARY       | PRIMARY | 8       | const | 24167 |   100.00 | Using where |
+----+-------------+------------------------------+------------+-------+---------------+---------+---------+-------+-------+----------+-------------+
1 row in set (0.00 sec)
```

There's substantially less checks if we know which specific ids we need to remove.

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -491,19 +491,47 @@ function xmldb_tool_excimer_upgrade($oldversion) {
     }
 
     if ($oldversion < 2024082301) {
-        // First we need to drop a few edge cases that have a length of 256.
-        $DB->delete_records_select('tool_excimer_page_groups', $DB->sql_length('name') . ' > 255');
+        $exists = $DB->record_exists_select('tool_excimer_page_groups', $DB->sql_length('name') . ' > 255');
+        if ($exists) {
+            // Change precision of name to 255 so it can be used as an index.
+            // First we need to drop a few edge cases that have a length of 256.
+            $DB->delete_records_select('tool_excimer_page_groups', $DB->sql_length('name') . ' > 255');
+        }
 
-        // SQL to delete duplicates while keeping the first occurrence.
-        $sql = "DELETE FROM {tool_excimer_page_groups}
+        // Check if there are any duplicates to be deleted by counting the
+        // expected total against the number of records in the table.
+        $expectedcount = $DB->count_records_sql('
+            SELECT count(*) FROM (
+                SELECT MIN(id) AS id
+                  FROM mdl_tool_excimer_page_groups
+                GROUP BY LOWER(name), month
+            ) a
+        ');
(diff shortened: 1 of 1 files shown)


jwalits:
Patch looks good to me

danmarsden:
Fyi - Looks like a hardcoded mdl prefix snuck into this pr.

PR #365, opened 2024-09-13
Add import and export profile functionality

### LINKED ISSUE #362: Export and import profiles
Mostly for safe keeping off site

bwalkerl:
Closes #362

### DIFF
--- a/classes/profile.php
+++ b/classes/profile.php
@@ -79,6 +79,21 @@ class profile extends persistent {
     /** Cron tasks */
     const SCRIPTTYPE_TASK = 4;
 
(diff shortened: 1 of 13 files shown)


classes/profile.php
```diff
@@ -62,6 +65,7 @@ class profile extends persistent {
         self::REASON_SLOW => 'slowest',
         self::REASON_FLAMEALL => 'flameall',
         self::REASON_STACK => 'stackdepth',
+        self::REASON_IMPORT => 'import',
```
brendanheywood, line 68: when exporting keep the existing reason and add in the import reason

PR #371, opened 2024-11-03
Migrate after_config hook callback

sumaiyamannan:
It should go into a new branch

PR #373, opened 2025-01-09
Add after_config callback for older Moodle versions

TomoTsuyuki:
The `core\hook\after_config hook` is for Moodle 4.5+, and 4.3 still needs the callback.

https://github.com/catalyst/moodle-tool_excimer/pull/371#issuecomment-2579160881

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -41,7 +41,7 @@ function tool_excimer_before_session_start() {
  * After config
  *
  * This is a legacy callback that is used for compatibility with older Moodle versions.
- * Moodle 4.5+ will use tool_abconfig\hook_callbacks::after_config instead.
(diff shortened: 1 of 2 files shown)


lib.php
```diff
@@ -37,6 +37,20 @@ function tool_excimer_before_session_start() {
     $manager = manager::get_instance();
 }
 
+/**
+ * After config
+ *
+ * This is a legacy callback that is used for compatibility with older Moodle versions.
+ * Moodle 4.5+ will use tool_abconfig\hook_callbacks::after_config instead.
```
dmitriim: leftover?

TomoTsuyuki: Good point.
I fixed the comment.

Issue #377, opened 2025-01-27
Some times excimer appears in its own profiles


brendanheywood:
![Image](https://github.com/user-attachments/assets/c64ef888-c7c1-44ad-9e86-e29ca0cad9b6)

It's not clear if this is genuinely profiling and finding itself, or if this is a weird display quirk. Assuming the worst that the bottleneck is excimer then some theories:
* something slow is happening which takes more than one excimer tick, eg a db insert
* excimer tries to do an insert of a partial record and gets a db lock or something slow which slows it down for longer than a tick
* the second excimer interrupt fires and the first excimer function is still being called and it snowballs

Another theory:
* the excimer manager is starting two profilers side by side and one ends up profiling the other
* it might not actually be the bottle neck but due to the timing synchronization its always running when the other one is running so the samples correlate

Some protections to add
* [ ] if we ever see excimer in the stack then put a hard halt to all processing during the process. We should still be able to persist it if needed in the shutdown handler
* [ ] try and reproduce running two profiles at the same time and if that is possible protect against that

Possibly related to https://github.com/catalyst/moodle-tool_excimer/issues/334

jaypha:
Do you have the profile data for the image above?

jaypha:
The Excimer profiler operates by interrupting the current PHP execution, doing it's thing, and then letting PHP execution resume. This is also true for the timer, which is used to run the partial saves.

The profiler is not switched off during a partial save, so if a 'tick' occurs during the partial save, then an entry will appear in the profile log. This should occur for certain if the save takes longer than the sampling period (e.g. >10ms).

This in itself should not cause problems. Excimer will simply save the stack and continue.

There are two possible ways in which something can go wrong that I can currently see.

One is that, if the partial save takes longer than the time interval (e.g. > 10 min), then the process function will be called even though the previous call has not finished. A simple check can be placed to prevent this.

The second is that, if the PHP was interrupted during an database access involving a lock, then the partial save will be blocked with no way of being unblocked. A freeze would occur because the other code has been stopped.

I don't know enough about database locking to be certain about this. But maybe we need to check to see if there is an existing lock when about to do a partial save.

PR #378, opened 2025-02-04
WR #452436: Prevent excimer processors from doubling up partial saves.

jaypha:
#377

### DIFF
--- a/classes/cron_processor.php
+++ b/classes/cron_processor.php
@@ -42,6 +42,9 @@ class cron_processor implements processor {
     /** @var int */
     protected $samplems;
 
+    /** @var bool */
+    protected static $alreadyprofiling = false;
+
     /**
      * Initialises the processor
      *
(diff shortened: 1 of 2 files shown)


classes/cron_processor.php
```diff
@@ -93,6 +96,14 @@ public function get_min_duration(): float {
      * @param manager $manager
      */
     public function on_interval(manager $manager) {
+        // We want to prevent doubling up of processing, so skip if an existing process is still executing.
+        // The profile logs will be kept and processed the next time.
+        if (self::$isinsideoninterval) {
```
brendanheywood: in cases when this has happened sometimes this is a very long running script, I would not want to have say 100s of the same debugging emitted in these cases I think just one call would be sufficient

classes/cron_processor.php
```diff
@@ -42,6 +42,9 @@ class cron_processor implements processor {
     /** @var int */
     protected $samplems;
 
+    /** @var bool */
+    protected static $isinsideoninterval = false;
```
brendanheywood: I find this a bit of a mouthful (eyefull??) to read, how about something simpler like $processing or $busy or $alreadyprofiling?

jaypha: fixed

classes/cron_processor.php
jaypha: fixed

jaypha: I also changed it so that it suspends future partial saves, as suggested in the WR.

brendanheywood (approved):

Issue #367, opened 2024-10-02
PHP message: PHP Warning:  rawurlencode() expects parameter 1 to be string, array given


x-iy:
PHP warning created when accessing webservice eg.

> /webservice/rest/server.php?wstoken=xxxxxxxxxxxxxx&wsfunction=core_course_get_courses&options[ids][0]=8804&moodlewsrestformat=json

> PHP message: PHP Warning:  rawurlencode() expects parameter 1 to be string, array given in /var/www/site/lib/weblib.php on line 522

Due to this line:
https://github.com/catalyst/moodle-tool_excimer/blob/MOODLE_35_STABLE/classes/script_metadata.php#L143

brendanheywood:
This has been worked around, the core issue still remains here:

https://tracker.moodle.org/browse/MDL-84668

PR #393, opened 2025-07-23
Fix handling of querystrings #367

bwalkerl:
Cherry-pick of #384 to the Totara 19 branch which also has the same issue.


djarran:
Looks good, same changes as commit that was cherry-picked

PR #394, opened 2025-07-29
purge_page_groups: make 'months ago' cutoff the middle of the month

SashaAnastasi:
Resolves month-rollover issue where previous months have fewer days.

Currently tool_excimer_purge_page_group_test fails on
* the 29th of March-September, unless it's a leap year;
* the 30th of March-September
* the 31st of every month that has a 31st

This is due to documented behaviour of the [php core function strtotime](https://www.php.net/manual/en/function.strtotime.php#refsect1-function.strtotime-notes).

### DIFF
--- a/classes/task/purge_page_groups.php
+++ b/classes/task/purge_page_groups.php
@@ -48,7 +48,7 @@ class purge_page_groups extends \core\task\scheduled_task {
         $months = get_config('tool_excimer', 'expiry_fuzzy_counts');
         // Only purge if a value is set.
         if (!empty($months)) {
-            $month = monthint::from_timestamp(strtotime(($months + 1) . ' months ago'));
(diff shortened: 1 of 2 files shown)


brendanheywood:
@SashaAnastasi is it possible to redo the unit withs with static data and checking those exact know time periods which have issues?

eg we have a method which takes a date as input, and we inject that date in the task and the unit tests. I find the original code somewhat confusing about its intent, but the new code is not any easier to understand why it is doing any of this date logic

SashaAnastasi:
@brendanheywood that's a good call, unfortunately I don't have capacity to implement it immediately.
If you could merge this fix for now to get the tests passing, I'll add issue #397 to my backlog.

bwalkerl:
Resolved by #400 which now uses static data

Issue #397, opened 2025-07-30
PHPUnit failed with: Failed asserting that 4 matches expected 5.


dmitriim:
Moodle 4.5 and 4.1

```
There was 1 failure:

1) tool_excimer\tool_excimer_purge_page_group_test::test_purge
Failed asserting that 4 matches expected 5.

/var/www/site/admin/tool/excimer/tests/tool_excimer_purge_page_group_test.php:80
/var/www/site/lib/phpunit/classes/advanced_testcase.php:80
phpvfscomposer:///var/www/site/vendor/phpunit/phpunit/phpunit:97
```

bwalkerl:
Should be resolved with #400

PR #400, opened 2025-08-25
Improve purge page group test and scheduled task

alexdamsted:
This work stemmed from a review of a pull request that tried to resolve this issue: https://github.com/catalyst/moodle-tool_excimer/issues/397

The review of that pull request is here: https://github.com/catalyst/moodle-tool_excimer/pull/394

This work also resolves the issue in https://github.com/catalyst/moodle-tool_excimer/issues/397. The difference is that it;

1. Tries to make the test and scheduled task more clear as to what they each do.
2. Introduces static month data and time injection to the test rather than using dynamic dates (safer and predictable).
3. Reduces time calculations in the affected test where possible. 
4. Add a new test to check if the user wants to retain more months than in the database, then keep all data. Seemed relevant to add, with time injection now possible in this test its easy to test edge cases like this.
5. The purge page groups scheduled task was also modified to suit the modifications made to the related test.

### DIFF
--- a/classes/task/purge_page_groups.php
+++ b/classes/task/purge_page_groups.php
@@ -39,17 +39,52 @@ class purge_page_groups extends \core\task\scheduled_task {
     }
 
     /**
-     * Do the job.
+     * Execute the scheduled task.
+     *
+     * @param int|null $now Optional timestamp for testing
      */
-    public function execute() {
+    public function execute(?int $now = null) {
+        $this->purge($now ?? time());
+    }
+
+    /**
+     * Purge page group records older than the retention period.
+     *
+     * @param int $now Current timestamp (injectable for tests)
+     */
+    public function purge(int $now): void {
         global $DB;
 
-        // Because we want to keep n full months of data, we add one to include the current month.
-        $months = get_config('tool_excimer', 'expiry_fuzzy_counts');
-        // Only purge if a value is set.
-        if (!empty($months)) {
-            $month = monthint::from_timestamp(strtotime(($months + 1) . ' months ago'));
-            $DB->delete_records_select(page_group::TABLE, 'month <= ' . $month);
+        $keepmonths = (int)get_config('tool_excimer', 'expiry_fuzzy_counts');
+
+        if (!empty($keepmonths) && $keepmonths > 0) {
+            $cutoff = $this->calculate_cutoff_month($now, $keepmonths);
+
+            // Delete all page group records less than or equal our cutoff month.
+            $DB->delete_records_select(page_group::TABLE, 'month <= ' . $cutoff);
         }
     }
+
+    /**
+     * Calculate the cutoff month number based on $now and months to keep.
+     *
+     * @param int $now Current timestamp
+     * @param int $keepmonths Number of months to retain
+     * @return int Cutoff month number (for comparison in DB)
+     */
+    public function calculate_cutoff_month(int $now, int $keepmonths): int {
+        // Because we want to keep n full months of data (n = $keepmonths).
+        // We add one month ($keepmonths += 1) to include the current month.
+        $keepmonths += 1;
+
+        $date = new \DateTime();
+        $date->setTimestamp($now);
+        $date->modify('first day of this month');
+        $date->modify("-{$keepmonths} months");
+
+        // Return a custom from_timestamp year/month number (e.g., 202508).
+        // We can call this custom timestamp the "cutoff month".
+        return monthint::from_timestamp($date->getTimestamp());
+    }
+
 }
--- a/tests/tool_excimer_purge_page_group_test.php
+++ b/tests/tool_excimer_purge_page_group_test.php
@@ -19,14 +19,14 @@ namespace tool_excimer;
 use tool_excimer\task\purge_page_groups;
 
 /**
- * <insertdescription>
+ * Testing the purge page group scheduled task execution method
  *
  * @package   tool_excimer
  * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
  * @copyright 2022, Catalyst IT
  * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
-class tool_excimer_purge_page_group_test extends \advanced_testcase {
+final class tool_excimer_purge_page_group_test extends \advanced_testcase {
 
     /**
      * Set up before each test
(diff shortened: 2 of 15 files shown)


tests/tool_excimer_purge_page_group_test.php
```diff
@@ -43,46 +43,76 @@ protected function setUp(): void {
      */
     public function test_purge() {
         global $DB;
-        $cutoff = 4;
-        $firstofthismonth = date('Y-m') . '-01';
+
+        // Set the number of months to retain
+        $keepmonths = 4;
+        set_config('expiry_fuzzy_counts', $keepmonths, 'tool_excimer');
```
bwalkerl: You shouldn't need to have set_config here as you're setting it to something else before you do anything.

classes/task/purge_page_groups.php
```diff
@@ -39,17 +39,52 @@ public function get_name() {
...
+     */
+    private function calculate_cutoff_month(int $now, int $keepmonths): int {
+        // Because we want to keep n full months of data (n = $keepmonths), we add one month ($keepmonths += 1) to include the current month
+        $keepmonths += 1;
+        
+        $date = new \DateTime();
+        $date->setTimestamp($now);
+        $date->modify('first day of this month');
+        $date->modify("-{$keepmonths} months");
+        $date->modify("second Monday of this month");
```
bwalkerl: I think we want to tighten the logic here. The other PR changing it to second Monday of the month was more an arbitrary change to get it to pass unit tests, but logically keeping it as the first is ideal (though we do want to fix the previous issues we had with DateTime)

tests/tool_excimer_purge_page_group_test.php
```diff
@@ -43,46 +43,76 @@ protected function setUp(): void {
...
-        $task->execute();
+        $task->execute($now->getTimestamp());
         $count = $DB->count_records(page_group::TABLE);
         $this->assertEquals(count($months), $count);
 
-        // Test a value of $cutoff, all but $cutoff + 1 rows should be purged.
-        set_config('expiry_fuzzy_counts', $cutoff, 'tool_excimer');
+        // Assertion 3: Keep more months than already exist, should do no purging
+        $count = $DB->count_records(page_group::TABLE);
+        if ($keepmonths > $count) {
```
bwalkerl: We shouldn't need these checks at all if the previous assertions pass?

tests/tool_excimer_purge_page_group_test.php
```diff
@@ -43,46 +43,76 @@ protected function setUp(): void {
...
-        $task->execute();
-
+        $task->execute($now->getTimestamp());
         $records = $DB->get_records(page_group::TABLE);
+        if (count($records) === $keepmonths) { 
+            $this->assertEquals($keepmonths, count($records));
+        } else if (count($records) > $keepmonths) {
+            // + 1 addition equation is here to ensure we keep the current month in addition to n full months (n = $keepmonths) because the scheduled task method calculate_cutoff_month requires we do that
+            $this->assertEquals($keepmonths + 1, count($records));
+        }
```
bwalkerl: Why is the check different depending on count? We should know how many records there are, so this should only need one assertion (the latter).

tests/tool_excimer_purge_page_group_test.php
```diff
@@ -43,46 +43,76 @@ protected function setUp(): void {
...
-
-        // Check that no month stored is earlier than the cutoff month.
-        $cutoffmonth = monthint::from_timestamp(strtotime(($cutoff + 1) . ' months ago'));
+        // Assertion 5: Ensure no month older than cutoff remains
+        // Use the same month calculation from the scheduled task method calculate_cutoff_month
+        $keepmonths+=1;
+        $now->modify('first day of this month');
+        $now->modify("-{$keepmonths} months");
+        $now->modify("second Monday of this month");        
+        $cutoffmonth = monthint::from_timestamp($now->getTimestamp());
```
bwalkerl: This would be better off making the function public and calling it directly.

tests/tool_excimer_purge_page_group_test.php
```diff
@@ -43,46 +43,76 @@ protected function setUp(): void {
...
         global $DB;
-        $cutoff = 4;
-        $firstofthismonth = date('Y-m') . '-01';
+
+        // Set the number of months to retain
+        $keepmonths = 4;
+        set_config('expiry_fuzzy_counts', $keepmonths, 'tool_excimer');
+
+        // Fixed "now" for an injectable static date test
+        $now = new \DateTime('2025-07-29');
```
bwalkerl: I think we should have the comment explain why we have chosen this date in particular (because the previous tests fail with it and it should pick up the issues with DateTime?). Maybe even with an assertion to prove it fails the old test.

bwalkerl (changes requested):
Thanks @alexdamsted 

This looks like it's on the right track, we just need to clean a few things up and tighten the logic.

tests/tool_excimer_purge_page_group_test.php
bwalkerl: Note that if we do that we would probably want some additional tests/assertions to test the function itself.

tests/tool_excimer_purge_page_group_test.php
alexdamsted: No extra full month to keep. An addition equation would fail here.

tests/tool_excimer_purge_page_group_test.php
alexdamsted: Removed unnecessary $count checks.

classes/task/purge_page_groups.php
alexdamsted: Removed unnecessary "second Monday of this month" logic.

tests/tool_excimer_purge_page_group_test.php
alexdamsted: Removed unnecessary set_config as shown.

tests/tool_excimer_purge_page_group_test.php
alexdamsted: Made the function public and used it directly in the test.

tests/tool_excimer_purge_page_group_test.php
bwalkerl: We should know both the number of records (because we set them, and because of asserts) and keepmonths - for the current unit tests there shouldn't be any that differ, and if there is it's a problem with the setup.

tests/tool_excimer_purge_page_group_test.php
alexdamsted: Changed the date to 31st July 2025. Added a comment to explain why this date was chosen. 

Assertion 4 covers the datetime rollover issue. Although, a more explicit assertion to test this would be nice to have.

tests/tool_excimer_purge_page_group_test.php
alexdamsted: Modified the check and fixed a test bug (subtraction needed instead of addition). Problem with the setup as suggested.

tests/tool_excimer_purge_page_group_test.php
bwalkerl: This still has the `$keepmonths > $count` check, which will never be met. What we need here is setting a different config number.

tests/tool_excimer_purge_page_group_test.php
bwalkerl: I think it should still be +1 and that the setup is still wrong.

tests/tool_excimer_purge_page_group_test.php
```diff
@@ -41,48 +41,86 @@ protected function setUp(): void {
...
-            monthint::from_timestamp(strtotime($firstofthismonth . '3 months ago')),
-            monthint::from_timestamp(strtotime($firstofthismonth . '4 months ago')),
-            monthint::from_timestamp(strtotime($firstofthismonth . '5 months ago')),
-            monthint::from_timestamp(strtotime($firstofthismonth . '6 months ago')),
-            monthint::from_timestamp(strtotime($firstofthismonth . '7 months ago')),
+            202507, // July 2025.
+            202506, // June 2025.
+            202505, // May 2025.
+            202204, // April 2022.
+            202103, // March 2021.
```
bwalkerl: There's 2022 and 2021 here instead of 2025?

tests/tool_excimer_purge_page_group_test.php
bwalkerl: I think it's still worth hardcoding an assert that $cutoff is the expected cutoff.

tests/tool_excimer_purge_page_group_test.php
bwalkerl: I think it's still worth hardcoding an assert that $cutoff is the expected cutoff, since it should be easy to calculate manually.

tests/tool_excimer_purge_page_group_test.php
alexdamsted: Added a different $keepmonths config number.

tests/tool_excimer_purge_page_group_test.php
alexdamsted: Changed to 2025 only.

tests/tool_excimer_purge_page_group_test.php
alexdamsted: Comment here spotted a bug with the test that made me think - 1 was the way forward: https://github.com/catalyst/moodle-tool_excimer/pull/400#discussion_r2297122625

Changed back to + 1.

tests/tool_excimer_purge_page_group_test.php
alexdamsted: Added an assert to make sure $cutoff is expected.

bwalkerl: I'm still not sold on the duplicated logic since if there were any issues both would have it. I just meant a hardcoded 202502 or whatever the accurate cutoff should be.

tests/tool_excimer_purge_page_group_test.php
bwalkerl: Thanks. That's working and the test is now being run - but the if statement is still redundant, as is the original `$keepmonths = 5`

alexdamsted: Removed the if statement and the original $keepmonths = 5.

tests/tool_excimer_purge_page_group_test.php
alexdamsted: Removed duplicated logic. Added a hard coded cutoff that is expected.

tests/tool_excimer_purge_page_group_test.php
```diff
@@ -41,48 +41,90 @@ protected function setUp(): void {
...
-        // Check that the number of records remaining is the correct number.
-        $this->assertEquals($cutoff + 1, count($records));
-
-        // Check that no month stored is earlier than the cutoff month.
-        $cutoffmonth = monthint::from_timestamp(strtotime(($cutoff + 1) . ' months ago'));
+        // Assertion 5: Ensure no month older than cutoff remains.
+        $cutoffmonth = $task->calculate_cutoff_month($now->getTimestamp(), $keepmonths);
         foreach ($records as $record) {
-            $this->assertGreaterThan($cutoffmonth, (int) $record->month);
+            $this->assertGreaterThanOrEqual($cutoffmonth, (int) $record->month);
```
bwalkerl: Is there any reason why this was changed to include Equal?

alexdamsted: This was changed in reaction to previous code which had bugs. Thanks for spotting this. Reverted this back.

bwalkerl (approved):

PR #401, opened 2025-08-26
Improve purge page group test and scheduled task

alexdamsted:
This work stemmed from a review of a pull request that tried to resolve this issue: https://github.com/catalyst/moodle-tool_excimer/issues/397

The review of that pull request is here: https://github.com/catalyst/moodle-tool_excimer/pull/394

This work also resolves the issue in https://github.com/catalyst/moodle-tool_excimer/issues/397. The difference is that it;

1. Tries to make the test and scheduled task more clear as to what they each do.
2. Introduces static month data and time injection to the test rather than using dynamic dates (safer and predictable).
3. Reduces time calculations in the affected test where possible.
4. Add a new test to check if the user wants to retain more months than in the database, then keep all data. Seemed relevant to add, with time injection now possible in this test its easy to test edge cases like this.
5. The purge page groups scheduled task was also modified to suit the modifications made to the related test.

### DIFF
--- a/classes/profile_helper.php
+++ b/classes/profile_helper.php
@@ -79,7 +79,7 @@ class profile_helper {
         $cachekey = $group;
         $cachefield = 'min_duration_for_reason_' . $reason . '_s';
         $cache = \cache::make('tool_excimer', 'request_metadata');
-        $result = $cache->get($cachekey) ?: array();
+        $result = $cache->get($cachekey) ?: [];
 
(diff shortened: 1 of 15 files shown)


bwalkerl (approved):
Thanks @alexdamsted 

I was wondering whether this branch supported void return types since they were added in PHP 7.1, but the readme lists the branch as PHP 7.1+ so it should be fine.

Issue #204, opened 2022-04-13
Serious bug: profile bottlenecks no longer working


brendanheywood:
Internal details:

https://etherpad.catalyst.net.nz/tool-excimer

![image](https://user-images.githubusercontent.com/187449/163092066-dd5472fe-95bc-4c2d-b6ea-41d1d685e70b.png)

PR #405, opened 2025-09-01
Fix samples re-grouping issue #403

### LINKED ISSUE #403: Samples are not always re-grouped to be more coarse
From time to time you see a profile like this:

2,066,272 samples @ ~10ms

In most cases the logic to gradually make the sampling time more coarse is working but clearly here it is not. There is a concern around holding a lot of memory to keep those 2 mil samples (this was a cron job that ran for 5 hours blocked on a single sql query). Perhaps this is only affecting the cron profiles

duyhuynhdev:
Fix samples re-grouping issue by removing event count and using real number of samples instead #403
Closes #403

### DIFF
--- a/classes/flamed3_node.php
+++ b/classes/flamed3_node.php
@@ -50,12 +50,10 @@ class flamed3_node {
      * Converts a trace tail to flamed3_nodes.
      *
      * @param array $tail
-     * @param int $eventcount the count of events
      */
-    public function add_excimer_trace_tail(array $tail, int $eventcount): void {
+    public function add_excimer_trace_tail(array $tail): void {
 
-        // Increases the base event size by the sum of its events.
-        $this->value += $eventcount;
+        $this->value += 1;
 
         if (count($tail)) {
             $child = end($this->children);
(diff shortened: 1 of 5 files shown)


classes/flamed3_node.php
```diff
@@ -50,12 +50,10 @@ public function __construct(string $name, int $value = 0, array $children = [])
...
      *
      * @param array $tail
-     * @param int $eventcount the count of events
      */
-    public function add_excimer_trace_tail(array $tail, int $eventcount): void {
+    public function add_excimer_trace_tail(array $tail): void {
 
-        // Increases the base event size by the sum of its events.
-        $this->value += $eventcount;
+        $this->value += 1;
```
bwalkerl, line 56: I'm concerned about the removal of this, because it was added to fix a bug that was regarded as serious https://github.com/catalyst/moodle-tool_excimer/issues/204 

Do we have any confirmation if your other changes resolves this issue as well? I can see the unit tests pass, but I'm not sure how well covered this is. As this changes internals I'm very hesitant to merge this without significant testing.

duyhuynhdev: The issue has been closed so I'll close that PR as well. Thanks

PR #417, opened 2025-12-03
Increase sampling time when reach sample limit #403

duyhuynhdev:
Refresh profiler period when we reach sample limit to slow down sampling speed.

Here is the test result:
```
Adhoc task id: 165
... started 02:36:24. Current memory use 20.2 MB.
Timer callback with log size 1000 at 1764729394
Timer callback with log size 1000 at 1764729404
Reach limit increase sampling time to 20 ms at 1764729404
Timer callback with log size 500 at 1764729414
Reach limit increase sampling time to 40 ms at 1764729414
Timer callback with log size 250 at 1764729424
Timer callback with log size 250 at 1764729434
Timer callback with log size 250 at 1764729444
Timer callback with log size 250 at 1764729454
Timer callback with log size 250 at 1764729464
Timer callback with log size 250 at 1764729474
Timer callback with log size 250 at 1764729484
Reach limit increase sampling time to 80 ms at 1764729484
Finish
```

### DIFF
--- a/classes/cron_processor.php
+++ b/classes/cron_processor.php
@@ -57,7 +57,7 @@ class cron_processor implements processor {
         });
 
         // Preload config values to avoid DB access during processing. See manager::get_altconnection() for more information.
-        $this->samplems = (int) get_config('tool_excimer', 'sample_ms');
+        $this->samplems = script_metadata::get_sampling_period();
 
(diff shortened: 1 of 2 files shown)


duyhuynhdev:
The issue has been closed so I'll close that PR as well. Thanks

Issue #418, opened 2025-12-04
Refine the doubling logic in sample set


duyhuynhdev:
Currently the doubling logic currently doesn't care how many events in a sample, as long as the sample's index is even it will be dropped. 
- For example:
  If we have sample list as below:
    0 - 10 events
    1 - 1000000 events
    2 - 10 events
    3 - 10 events
  If we want to half that right now the logic would end up with 
    0 - 10 events
    2 - 10- events

The important sample and a significant of event has been removed and its could not show the bottleneck (see #204)

So that the logic for doubling should be changed. Instead of dropping every second row we can merge them instead by

1. Adding the event counts
For example :
  0 - 10 events
  1 - 1000000 events
Would become
  1 - 1000010 events

2. Picking the stack of the one which is bigger

duyhuynhdev:
Another idea from @brendanheywood , see below:

0 - 10 events A
1 - 100 events B
2 - 100 events C
3 - 10 events D
4 - 10 events E
5 - 10 events F

if we halve from 6 to 3 then we break it into 3 groups of events. The total event count is 240 so we want 3 groups of 80

0 - 10 events A
1 - 70 events B

1 - 30 events B
2 - 50 events C

2 - 50 events C
3 - 10 events D
4 - 10 events E
5 - 10 events F


then for each group we combine and pick the large event
0 - 80 events B

1 - 80 events C

2 - 80 events C
there will be rounding errors no matter what we do

PR #420, opened 2025-12-10
Refine the doubling logic #418

duyhuynhdev:
- Update the doubling logic according to the first approach, as the second approach still has corner cases where a single log event can dominate others if it has a large event count.
- Remove the timer callback and use the profiler flush callback instead. In addition, expose the new maxsample setting to support the flush callback.
- Update the relevant classes and unit tests to align with the new workflow and doubling logic.
- Add new unit tests to cover the updated behaviour.
- Fix CI complaints, including unsorted language files.


classes/cron_processor.php
```diff
@@ -39,11 +39,23 @@ class cron_processor implements processor {
...
+
+    /** @var int */
+    protected $logcount = 0;
 
     /** @var bool */
     protected static $alreadyprofiling = false;
 
+    /** @var array */
+    protected static $logs = [];
+
```
bwalkerl: There's a lot of shared code between cron and web processors, particularly with what has been added. Is there a reason why there isn't a child class?

classes/cron_processor.php
```diff
@@ -52,25 +64,42 @@ class cron_processor implements processor {
...
-        });
-
         // Preload config values to avoid DB access during processing. See manager::get_altconnection() for more information.
-        $this->samplems = (int) get_config('tool_excimer', 'sample_ms');
+        $this->samplingperiod = script_metadata::get_sampling_period();
+        $this->samplelimit = script_metadata::get_sample_limit();
+        $this->maxsamples = script_metadata::get_max_samples();
+        $manager->get_profiler()->setFlushCallback(function ($log) use ($manager) {
+            $this->on_interval($log, $manager);
+        }, $this->maxsamples);
```
bwalkerl, line 55: One major difference here is that from your changes setFlushCallback seems to be called on destruction. Is this the case? If so it would be worthwhile having a comment describing this.

classes/cron_processor.php
```diff
@@ -52,25 +64,42 @@ class cron_processor implements processor {
...
+
+    /**
+     * Get sample limit of processor
+     *
+     * @return int sample_limit
+     */
+    public function get_sample_limit() {
+        return $this->samplelimit;
+    }
+
```
bwalkerl, line 84: These are unused?

classes/cron_processor.php
```diff
@@ -92,65 +133,71 @@ public function get_min_duration(): float {
...
             debugging('tool_excimer: starting cron_processor::on_interval when previous call has not yet finished');
-            if ($isfinal) {
+            // The final flush call when profiler is destroyed.
+            if ($log->count() < $this->samplelimit) {
                 // This should never happen.
                 debugging('tool_excimer: alreadyprofiling is true during final on_interval.');
             }
             return;
         }
         self::$alreadyprofiling = true;
```
bwalkerl, line 118: The logic to prevent double up of processing may not be needed anymore now that we've switched from timers to flush callbacks (I'm not sure if this is paused during the callbacks). I think it's worth looking into further because it would be great if we could remove $self::logs entirely.

classes/cron_processor.php
```diff
@@ -92,65 +133,71 @@ public function get_min_duration(): float {
...
+        foreach (self::$logs as $log) {
+            $memoryusage = memory_get_usage();
```
bwalkerl, line 120: Processing self::$logs like this won't have accurate numbers for memory usage and a few other things at the time of processing. Would be good if we can get rid of it altogether as per the previous comment.

classes/cron_processor.php
```diff
@@ -92,65 +133,71 @@ public function get_min_duration(): float {
...
+                $taskname = $this->findtaskname($sample);
+                $sampletime = $manager->get_starttime() + $sample->getTimestamp();
+
+                // If there is a task and the current task name is different from the previous, then store the profile.
+                if ($this->tasksampleset && ($this->tasksampleset->name != $taskname)) {
+                    $profile = $this->process($manager, $this->sampletime);
+                    // Keep an approximate count of each profile.
+                    page_group::record_fuzzy_counts($profile);
+                    $this->tasksampleset = null;
+                }
```
bwalkerl, line 131: I think detecting new tasks like this is incompatible with the increased sample periods. Because of this, new tasks will start off with the increased sample periods from the previous tasks and likely be missing the most important parts of the profile. 

This is likely the biggest issues here and I'm not sure the best way of handling this.

classes/flamed3_node.php
```diff
@@ -69,18 +69,16 @@ public function add_excimer_trace_tail(array $tail, int $eventcount): void {
...
-     * Extracts data from Excimer log entries and converts to a flame node tree, compatible with
+     * Extracts data from sample set samples and converts to a flame node tree, compatible with
      * d3-flame-graph.
      *
-     * @param iterable $entries
+     * @param array $samples samples
      * @return flamed3_node
      */
-    public static function from_excimer_log_entries(iterable $entries): flamed3_node {
+    public static function from_sample_set_samples(array $samples): flamed3_node {
```
bwalkerl, line 78: I think this should stay as iterable to match excimer.

classes/sample_set.php
```diff
@@ -75,27 +69,29 @@ public function get_stack_depth(): int {
...
             }
-            return;
+            $this->samples[] = [
+                'eventcount' => $eventcount,
+                'trace' => $trace,
+            ];
+        } else {
+            $this->samples[] = $sample;
+            $this->totaladded++;
+            $ismemory = true;
```
bwalkerl, line 93: Is this always memory or can there be other edge cases that aren't \ExcimerLogEntry? Worth a comment.

classes/sample_set.php
```diff
@@ -110,24 +106,55 @@ public function add_many_samples(iterable $samples) {
...
+    private function merge_memory_usage_sample_set() {
+        $newsamples = [];
+        for ($i = 0; $i < count($this->samples); $i += 2) {
+            // Prevent sample limit is odd.
+            if ($i == count($this->samples) - 1) {
+                $newsamples[] = $this->samples[$i];
+            } else {
+                $newsamples[] = [
+                    'sampleindex' => $this->samples[$i]['sampleindex'],
+                    'value' => ($this->samples[$i]['value'] + $this->samples[$i + 1]['value']) / 2.0,
```
bwalkerl: Would max memory be more useful than average here? Would also avoid issues with floats. Also worth having some comments.

classes/sample_set.php
```diff
@@ -110,24 +106,55 @@ public function add_many_samples(iterable $samples) {
...
+                } else {
+                    $trace = $this->samples[$i + 1]['trace'];
+                }
+                $newsamples[] = [
+                    'eventcount' => ($this->samples[$i]['eventcount'] + $this->samples[$i + 1]['eventcount']),
+                    'trace' => $trace,
+                ];
+            }
+        }
+        $this->samples = $newsamples;
```
bwalkerl, line 167: Would be useful to have comments explaining why we're doing this.

classes/sample_set.php
```diff
@@ -75,27 +69,29 @@ public function get_stack_depth(): int {
...
             $trace = $sample->getTrace();
+            $this->totaladded += $eventcount;
             if ($trace) {
                 $this->maxstackdepth = max($this->maxstackdepth, count($trace));
             }
-            return;
+            $this->samples[] = [
+                'eventcount' => $eventcount,
+                'trace' => $trace,
+            ];
```
bwalkerl, line 89: What other information can $sample contain? We could be losing some data that is useful. 

Ideally we keep these as a $sample after merging, but if that's not possible would need to update the docs of $samples.

classes/web_processor.php
```diff
@@ -109,51 +119,75 @@ public function get_min_duration(): float {
...
+     * @param \ExcimerLog $log
      * @param manager $manager
      * @param bool $isfinal
      * @throws \dml_exception
      */
-    public function process(manager $manager, bool $isfinal) {
+    public function process($log, manager $manager, bool $isfinal = false) {
         // We want to prevent overlapping of processing, so skip if an existing process is still executing.
         // The profile logs will be kept and processed the next time.
+        self::$logs[] = $log;
```
bwalkerl, line 110: See comments about overlapping processing in cron.

classes/web_processor.php
```diff
@@ -109,51 +119,75 @@ public function get_min_duration(): float {
...
+
+        // Doubling sampling period if it reaches the limit.
+        $this->logcount += $log->count();
+        if ($this->partialsave && $this->logcount >= $this->samplelimit) {
```
bwalkerl, line 113: Do we want this applied to non-partial saves too? I believe it was previously applied there as well.

bwalkerl (changes requested):
Thanks @duyhuynhdev 

This is a big change so we're going to need to test this thoroughly. I've started with comments from a code review, but I haven't tested this personally yet. 

My main concerns so far is how this affects task detection and whether we are losing any sample data when we change the structure while combining components.

classes/cron_processor.php
duyhuynhdev: Yes, the flush callback will be called on destruction. I have added the comment in the next commit

classes/cron_processor.php
duyhuynhdev: They are being used in UTs for checking changes when on_reach_limit is called

classes/web_processor.php
```diff
@@ -78,23 +85,26 @@ public function init(manager $manager) {
...
         if ($this->partialsave) {
-            $manager->get_timer()->setCallback(function () use ($manager) {
+            $manager->get_profiler()->setFlushCallback(function ($log) use ($manager) {
                 // Once overlapping has happened once, we prevent all future partial saving.
                 if (!$this->hasoverlapped) {
-                    $this->process($manager, false);
+                    $this->process($log, $manager);
                 }
-            });
+            }, $this->maxsamples);
```
bwalkerl: With this change partial save never marks the profile as finished. 

I'm also thinking that with this patch it might be OK to re-enable partial saves by default (and lighten the old warnings) as the number of samples won't rise quickly when the DB is having issues, which was the main reason we disabled it..

classes/sample_set.php
```diff
@@ -110,24 +106,55 @@ public function add_many_samples(iterable $samples) {
...
+            if ($i == count($this->samples) - 1) {
+                $newsamples[] = $this->samples[$i];
+            } else {
+                if ($this->samples[$i]['eventcount'] >= $this->samples[$i + 1]['eventcount']) {
+                    $trace = $this->samples[$i]['trace'];
+                } else {
+                    $trace = $this->samples[$i + 1]['trace'];
+                }
+                $newsamples[] = [
+                    'eventcount' => ($this->samples[$i]['eventcount'] + $this->samples[$i + 1]['eventcount']),
```
bwalkerl: With the change in event counts the hover duration shown on the graph is incorrect after merging and needs to be updated.

lang/en/tool_excimer.php
```diff
@@ -60,192 +59,152 @@
...
 $string['field_maxlockheld'] = 'Longest';
+$string['field_memoryusagemax'] = 'Max Memory Used';
+$string['field_mincreated'] = 'Earliest';
+$string['field_minduration'] = 'Fastest';
 $string['field_minlockheld'] = 'Shortest';
-$string['field_requestcount'] = 'Num profiles';
+$string['field_month'] = 'Month';
+$string['field_name'] = 'Name';
+$string['field_numsamples'] = 'Number of samples';
+$string['field_numsamples_value'] = '{$a->samples} samples ({$a->events} events) @ ~{$a->samplerate}ms';
```
bwalkerl, line 103: We need to make this consistent with the hover display on the graph, which currently only uses 'samples' for events. We probably need to iron out the terminology since samples will have more meaning than events to most ends users. 

I also think it would be better to keep the old display without events when the number of events is 0 (old profiles etc)

bwalkerl (commented):
I have a couple more comments from testing. The transformations of the graphs after merging also doesn't look right to me (tested with partial saves to see before and after), so will need to test that more.

classes/cron_processor.php
duyhuynhdev: At this stage, the exact underlying scenario isn’t clear because the documentation doesn’t cover this case. However, this likely happens when maxSamples is set too low, causing the process to continue running while callbacks are triggered repeatedly. In this scenario, self::logs can be used for stored new coming samples.

classes/cron_processor.php
duyhuynhdev: I'm keeping this logic as it was before. As mentioned above, removing the double processing prevention would be risky. I could sampling a memory and attach it along with the samples `log` into `self::$logs` if you want to make it more precise. Note that this only applies to the duplicate handling case. In normal cases, samples are added to the  self::$logs but then retrieved and processed immediately, so I think it wouldn't affect this case.

classes/sample_set.php
duyhuynhdev: I agree, max should be better

classes/cron_processor.php
bwalkerl: Previously there were edge cases of the 10s interval overlapping, see #377 

With your changes these should be much rarer as in the worst case the number of possible samples in on_interval should be very low. I'm not even sure if the profiler would be taking samples here.

I'm interested in what happens when max samples is set too low. The previous issue this fixed shouldn't be a concern if it only happens with bad config, but is there a chance of endless loops if this is set to something ridiculous like 1?

classes/web_processor.php
duyhuynhdev: We only get sampling once with non non-partial saves which means we won't set the callback for non-partial saves ( as previous) so we the sampling period doubling cannot be applied this case. However, the merge is still applied when we process the samples.

classes/sample_set.php
duyhuynhdev: There are only 2 type of samples:
- A task sample retrieved from ExcimerLogEntry
- A memory usage sample created from our code

For the first one , except the event count and trace , ExcimerLogEntry also has timestamp which is the floating point number of seconds since the ExcimerProfiler object was constructed. However, we will not use it anywhere so I won't collect it.
https://github.com/wikimedia/php-excimer/blob/master/stubs/ExcimerLogEntry.php

classes/cron_processor.php
bwalkerl: Agreed that this is the same as before and not a huge issue. I just think it's adding unnecessary complexity and invalid data points if it's no longer needed.

classes/web_processor.php
duyhuynhdev: Yeah, it is a bug. Let me fix it

classes/web_processor.php
bwalkerl: Thanks, that clarifies things and looks correct.

In that case, what happens if partial save is enabled and this is called by the flush callback when the ExcimerProfiler object is destroyed? Restarting it there seems bad.

classes/sample_set.php
bwalkerl: Thanks, unfortunately the \ExcimerLogEntry class has `private final function __construct()` so we can't use that, so we are stuck with arrays or a new class. It's a bit messy, especially with memory using the same, but it works. 

Might be worth linking the docs here.

classes/processor.php
```diff
@@ -33,18 +33,62 @@
  *
  * @package   tool_excimer
  */
-interface processor {
+class processor {
```
bwalkerl: I'd recommend making this an abstract class and keeping init as an abstract method

classes/sample_set.php
```diff
@@ -110,24 +106,55 @@ public function add_many_samples(iterable $samples) {
...
+            if ($i == count($this->samples) - 1) {
+                $newsamples[] = $this->samples[$i];
+            } else {
+                if ($this->samples[$i]['eventcount'] >= $this->samples[$i + 1]['eventcount']) {
+                    $trace = $this->samples[$i]['trace'];
+                } else {
+                    $trace = $this->samples[$i + 1]['trace'];
+                }
+                $newsamples[] = [
+                    'eventcount' => ceil(($this->samples[$i]['eventcount'] + $this->samples[$i + 1]['eventcount']) / 2),
```
bwalkerl: Is ceiling preferred here? Always rounding up will increase the margin of error. The number of samples won't be perfect unless we allow decimals, but maybe we can introduce some logic here to keep it more balanced.

I also still think we need more comments about what we're doing here.

classes/web_processor.php
```diff
@@ -62,102 +50,118 @@ public function __construct() {
...
+     *
+     * We then check for a new task with the current sample.
      *
+     * @param \ExcimerLog $log
      * @param manager $manager
-     * @param bool $isfinal
-     * @throws \dml_exception
      */
-    public function process(manager $manager, bool $isfinal) {
+    public function on_interval($log, manager $manager) {
```
bwalkerl: Technically, on_interval should be renamed (in both classes) as we're no longer using intervals.

bwalkerl (commented):
Thanks for the updates @duyhuynhdev 

I've left a couple more comments, plus responses to some of your other comments that have actions.

classes/web_processor.php
bwalkerl: After the update partial save is no longer saving partial profiles every time there is a flush callback.

classes/cron_processor.php
duyhuynhdev: Yes, it should be possible, because we don’t have any constraint on the maxSamples value. That’s why I introduced the $logs array as a waiting list. New logs from the callback are collected in this array until the current process completes. I’m not sure how often overlaps will occur with the new logic, but $logs won’t impact the existing flow. On the contrary, it helps mitigate overlapping if it does occur.

classes/cron_processor.php
duyhuynhdev: In my view, the sampling period isn’t tied to a specific task. It’s increased to reduce impact on site performance, as stated. So even if there’s a new task, if the site is very busy, a higher sampling period is still reasonable.

Regarding missing information: if a task is lightweight and completes before the next sampling interval, I think that’s acceptable. Our focus is on bottlenecks (heavy tasks). Lightweight tasks are also filtered out, at least at the stack trace level, by the merge logic.

classes/sample_set.php
duyhuynhdev: According to the code, we have only 2 types of the sample set: a tasksampleset (ExcimerLogEntry) and a memoryusagesampleset. So there is no other edge cases that aren't \ExcimerLogEntry

classes/sample_set.php
duyhuynhdev: Thanks I will add a description there

classes/sample_set.php
duyhuynhdev: I will update the docs and explaination

classes/cron_processor.php
bwalkerl: This is a serious regression, and is much more broken than you think.

Multiple cron tasks are processed in a single call. The sampling period is never reset. A task that starts several hours in would have a ridiculously high sampling period and not contain the samples required to debug it properly.

Even if you reset that when a new task is detected, the gap of processing at higher intervals is likely to cause issues following long tasks. On top of this going from a higher sampling period to a lower sampling period in the middle of the task would significantly distort the display and misrepresent parts of the task.

classes/web_processor.php
duyhuynhdev: $this->logcount is used to manage increases to the samplingperiod.Even in the final flush callback, the only affected element is the samplingperiod. So, it should not impact any profile, because the samplerate is no longer derived from the samplingperiod.

classes/sample_set.php
bwalkerl: That was my understanding as well, but this should be made clear with a comment.

classes/web_processor.php
duyhuynhdev: I will update it in the next commit

classes/web_processor.php
bwalkerl: I'm less concerned about the profile and more about the harm of restarting the profiler while it's being destroyed. It could be handled gracefully behind the scenes, but we should be able to detect this ourselves and explicitly handle the logic for not restarting it in this case.

classes/sample_set.php
duyhuynhdev: It has been updated from summing to halving as discussed

lang/en/tool_excimer.php
duyhuynhdev: number of events should not be 0. If there is no `numevents` it will use the numsamples instead. 
```
'events' => number_format($data['numevents'] ?? $data['numsamples'], 0, $decsep, $thousandssep),
```

classes/web_processor.php
duyhuynhdev: I renamed it to on_flush

classes/cron_processor.php
duyhuynhdev: I think we need specific cases to evaluate this. It’s too abstract for me to visualize what’s happening behind the scenes. Based on what you’re saying, it sounds like we should not be updating the sampling period.

lang/en/tool_excimer.php
bwalkerl: Old profiles prior to the upgrade are showing as 0 in the database to me, but this isn't a big concern as old profiles will be flushed out eventually. That main issue here is making the terminology consistent with the graph.

classes/cron_processor.php
bwalkerl: Yes, I feel like this is the main reason cron was processed on interval before. This did come with the drawback of taking extra samples and then discarding them based on the filter it, so there's still value in swapping if we can handle this somehow.

duyhuynhdev: I’m happy to remove the sampling-period doubling and keep the flush callback, but I can’t justify the mismatch between what we are trying to solve and what we’re doing. At this point, the sample limit is primarily to prevent excessive log storage, rather than to address sampling spam during long-running tasks.

classes/web_processor.php
duyhuynhdev: Thanks, I separated the on_flush and profile process functions.

classes/sample_set.php
duyhuynhdev: Sorry. I've updated it to the round() function. I intended to use the round function but somehow used the ceil() function instead.

classes/processor.php
duyhuynhdev: Thanks I have updated

classes/sample_set.php
bwalkerl: Round makes more sense, but it's going to run into the same problems as it will always be .5.

duyhuynhdev: Is rounding to the nearest 0.5 a problem? For example, rounding 3.5 to 4.0 or 3.4 to 3.0 is acceptable to me, because the small difference should not materially affect the analysis.

bwalkerl: By itself it's not an issue, but since this will always be rounding up the number of samples is going to be slightly overestimated which can have flow on effects to duration estimates. 

I'm not sure how much of a problem this is in practice - if there's only one or two with more samples it's not an issue, but if say a tenth of the samples have this then it could have a larger impact.

classes/web_processor.php
bwalkerl: Now it's updating (or rather, will once you fix the typo ')'), but it will save the last profile twice. We want to avoid this redundancy - we shouldn't be adding extra DB updates when they're not needed.

duyhuynhdev: Hey Ben, can you explain more about the typo and saving the last profile twice issue?

bwalkerl:
Closing this in favour of https://github.com/catalyst/moodle-tool_excimer/pull/423 - This PR does still have some different code that can be reused in the future.

PR #423, opened 2026-01-30
Refine the doubling logic for web #418

duyhuynhdev:
#418

Add new doubling logic for web processor only
Update related classes and tests

### DIFF
--- a/classes/cron_processor.php
+++ b/classes/cron_processor.php
@@ -32,10 +32,10 @@ class cron_processor implements processor {
     /** @var float Timestamp updated after processing each sample */
     public $sampletime;
 
-    /** @var sample_set A sample set recorded while processing a task */
+    /** @var excimer_sample_set A sample set recorded while processing a task */
     public $tasksampleset = null;
 
-    /** @var sample_set A sample set for memory usage recorded while processing a task */
+    /** @var memory_sample_set A sample set for memory usage recorded while processing a task */
     protected $memoryusagesampleset;
 
     /** @var int */
@@ -126,8 +126,8 @@ class cron_processor implements processor {
 
             // If there exists a current task, and the sampleset for it is not created yet, create it.
             if ($taskname && ($this->tasksampleset === null)) {
-                $this->tasksampleset = new sample_set($taskname, $this->sampletime);
-                $this->memoryusagesampleset = new sample_set($taskname, $this->sampletime);
+                $this->tasksampleset = new excimer_sample_set($taskname, $this->sampletime);
+                $this->memoryusagesampleset = new memory_sample_set($taskname, $this->sampletime);
                 if ($memoryusage) { // Ensure this only adds the mem usage for the initial base sample due to accuracy.
                     $this->memoryusagesampleset->add_sample(['sampleindex' => 0, 'value' => $memoryusage]);
                     $memoryusage = 0;
@@ -196,7 +196,8 @@ class cron_processor implements processor {
             $profile->set('reason', $reasons);
             $profile->set('finished', (int) $finishtime);
             $profile->set('memoryusagedatad3', $this->memoryusagesampleset->samples);
-            $profile->set('flamedatad3', flamed3_node::from_excimer_log_entries($this->tasksampleset->samples));
+            $profile->set('flamedatad3', flamed3_node::from_sample_set_samples($this->tasksampleset->samples));
+            $profile->set('numlogentries', count($this->tasksampleset->samples));
             $profile->set('numsamples', $this->tasksampleset->count());
             $profile->set('samplerate', $this->tasksampleset->filter_rate() * $this->samplems);
             $profile->save_record();
--- a/classes/flamed3_node.php
+++ b/classes/flamed3_node.php
@@ -49,9 +49,9 @@ class flamed3_node {
      * Converts a trace tail to flamed3_nodes.
      *
      * @param array $tail
-     * @param int $eventcount the count of events
+     * @param float $eventcount the count of events
      */
-    public function add_excimer_trace_tail(array $tail, int $eventcount): void {
+    public function add_excimer_trace_tail(array $tail, float $eventcount): void {
 
         // Increases the base event size by the sum of its events.
         $this->value += $eventcount;
@@ -69,18 +69,16 @@ class flamed3_node {
     }
 
     /**
-     * Extracts data from Excimer log entries and converts to a flame node tree, compatible with
+     * Extracts data from sample set samples and converts to a flame node tree, compatible with
      * d3-flame-graph.
      *
-     * @param iterable $entries
+     * @param array $samples samples
      * @return flamed3_node
      */
-    public static function from_excimer_log_entries(iterable $entries): flamed3_node {
+    public static function from_sample_set_samples(array $samples): flamed3_node {
         $root = new flamed3_node('root');
-        foreach ($entries as $entry) {
-            $eventcount = $entry->getEventCount();
-            $trace = array_reverse($entry->getTrace());
-            $root->add_excimer_trace_tail($trace, $eventcount);
+        foreach ($samples as $sample) {
+            $root->add_excimer_trace_tail(array_reverse($sample['trace']), $sample['eventcount']);
         }
         return $root;
     }
--- a/classes/sample_set.php
+++ b/classes/sample_set.php
@@ -24,28 +24,26 @@ namespace tool_excimer;
  * @copyright  2022 Catalyst IT
  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
-class sample_set {
+abstract class sample_set {
     /** @var string name of the sample set. */
     public $name;
     /** @var float Starting time of the sample set. */
     public $starttime;
 
-    /** @var array An array of \ExcimerLogEntry objects. */
+    /** @var array An array of sample objects. It could contain memory usage samples or tasks samples  */
     public $samples = [];
+
+    /** @var array An array of sample objects. It is used to store sample upto the filterrate  */
+    public $samplepool = [];
 
     /** @var int Sample limit. */
     public $samplelimit;
-    /** @var int The maximum stack depth. */
-    public $maxstackdepth = 0;
 
     /** @var int If is R, then only each Rth sample is recorded. */
-    private $filterrate = 1;
-
-    /** @var int Internal counter to help with filtering. */
-    private $counter = 0;
+    protected $filterrate = 1;
 
     /** @var int Internal counter of how many samples were added (regardless of how many are currently held). */
-    private $totaladded = 0;
+    protected $totaladded = 0;
 
     /**
      * Constructs the sample set.
@@ -61,41 +59,42 @@ class sample_set {
     }
 
     /**
-     * Return the stack depth for this set.
-     *
-     * @return int
-     */
-    public function get_stack_depth(): int {
-        return (int) $this->maxstackdepth;
-    }
-
-    /**
      * Add a sample to the sample store, applying any filters.
      *
      * @param array|\ExcimerLogEntry $sample
      */
     public function add_sample($sample) {
-        $trace = false;
         if (count($this->samples) === $this->samplelimit) {
             $this->apply_doubling();
         }
-        $this->counter += 1;
-        if ($this->counter === $this->filterrate) {
-            $this->samples[] = $sample;
-            $this->counter = 0;
+        $this->process_samples($sample);
+        if (count($this->samplepool) === $this->filterrate) {
+            $this->samples[] = $this->get_sample_from_pool();
         }
-        // If this is a log entry, it will count the number of total events
-        // processed instead.
-        // Each time a sample is added, recalculate the maxstackdepth for this set.
-        if ($sample instanceof \ExcimerLogEntry) {
-            $this->totaladded += $sample->getEventCount();
-            $trace = $sample->getTrace();
-            if ($trace) {
-                $this->maxstackdepth = max($this->maxstackdepth, count($trace));
-            }
-            return;
-        }
-        $this->totaladded++;
+    }
+
+    /**
+     * Processing given samples before adding into the sample set.
+     *
+     * @param array|\ExcimerLogEntry $sample
+     */
+    abstract public function process_samples($sample);
+
+    /**
+     * Merging sample set
+     *
+     * @param array $samples
+     * @param int $chunksize
+     */
+    abstract public function merge_sample_set(array $samples, int $chunksize = 2);
+
+    /**
+     * Get a significant sample from pool.
+     */
+    private function get_sample_from_pool(): array {
+        $mergedsample = $this->merge_sample_set($this->samplepool, $this->filterrate);
+        $this->samplepool = [];
+        return reset($mergedsample);
     }
 
     /**
@@ -110,24 +109,14 @@ class sample_set {
     }
 
     /**
-     * Doubles the filter rate, and strips every second sample from the set.
+     * Merge samples to increase storage
      * Called when the sample limit is reached.
-     *
-     * We have two options here. Either double the sampling period, or apply a filter to record only the
-     * Nth sample passed to sample_set. By using a filter and keeping the sampling period the same, we avoid
-     * spilling over into the next task.
      */
-    public function apply_doubling() {
+    public function apply_doubling(): void {
         $this->filterrate *= 2;
-        $this->samples = array_values(
-            array_filter(
-                $this->samples,
-                function ($key) {
-                    return ($key % 2);
-                },
-                ARRAY_FILTER_USE_KEY
-            )
-        );
+        // Instead of dropping a half of samples, we merge them together to keep useful information.
+        // The merge logic is different for each type of sample set.
+        $this->samples = $this->merge_sample_set($this->samples);
     }
 
     /**
@@ -135,7 +124,7 @@ class sample_set {
      *
      * @return int number of samples added
      */
-    public function total_added() {
+    public function total_added(): int {
         return $this->totaladded;
     }
 
@@ -147,16 +136,7 @@ class sample_set {
      *
      * @return int count of $this->samples
      */
-    public function count(): int {
-        $count = count($this->samples);
-        if ($count > 0 && $this->samples[0] instanceof \ExcimerLogEntry) {
-            $count = array_reduce($this->samples, function ($acc, $sample) {
-                $acc += $sample->getEventCount();
-                return $acc;
-            }, 0);
-        }
-        return $count;
-    }
+    abstract public function count(): int;
 
     /**
      * Returns the filter rate to calculate the real sampling rate
--- a/classes/web_processor.php
+++ b/classes/web_processor.php
@@ -30,9 +30,9 @@ class web_processor implements processor {
     /** @var profile The profile object for the run. */
     protected $profile;
 
-    /** @var sample_set */
+    /** @var excimer_sample_set */
     protected $sampleset;
-    /** @var sample_set */
+    /** @var memory_sample_set */
     protected $memoryusagesampleset;
 
     /** @var int */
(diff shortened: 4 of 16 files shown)


classes/sample_set.php
```diff
@@ -75,27 +75,53 @@ public function get_stack_depth(): int {
...
+    }
+
+    /**
+     * Get a significant sample from pool.
+     *
+     * @param bool $ismemory
+     */
+    private function get_sample_from_pool(bool $ismemory = false): array {
+        if ($ismemory) {
+            $mergedsample = $this->merge_memory_usage_sample_set($this->samplepool, count($this->samplepool));
```
bwalkerl: I think passing the filterrate instead of a count of the samplepool would make the code easier to understand.

classes/sample_set.php
```diff
@@ -75,27 +75,53 @@ public function get_stack_depth(): int {
...
+        } else {
+            $mergedsample = $this->merge_excimer_sample_set($this->samplepool, count($this->samplepool));
+        }
+        if (count($mergedsample) > 1) {
+            debugging('tool_excimer: get_sample_from_pool is running incorrectly.');
```
bwalkerl: How likely are we every going to hit this condition? I think we should remove this and make sure it's covered in unit tests instead.

classes/sample_set.php
```diff
@@ -30,9 +30,12 @@ class sample_set {
     /** @var float Starting time of the sample set. */
     public $starttime;
 
-    /** @var array An array of \ExcimerLogEntry objects. */
+    /** @var array An array of sample objects. It could contain memory usage samples or tasks samples  */
     public $samples = [];
 
+    /** @var array An array of sample objects. It could contain memory usage samples or tasks samples  */
+    public $samplepool = [];
```
bwalkerl: The phpdocs here shouldn't be the same as $samples, it should describe the purpose more.

classes/sample_set.php
```diff
@@ -75,27 +75,53 @@ public function get_stack_depth(): int {
...
+            $this->samplepool[] = $sample;
+            $this->totaladded++;
+            $ismemory = true;
+        }
+        if (count($this->samplepool) === $this->filterrate) {
+            $this->samples[] = $this->get_sample_from_pool();
+        }
+        if (count($this->samples) === $this->samplelimit) {
+            $this->apply_doubling($ismemory);
+        }
```
bwalkerl: Previously doubling was applied before the new sample was added. Is there any benefit from changing the order?

lang/en/tool_excimer.php
```diff
@@ -138,6 +138,9 @@
     It will not be purged during cleanup tasks, nor can it be deleted manually (will also be excluded from group deletes).
     Typically you would provide a reason why you want to keep this profile. Clearing this box will allow the profile to be deleted.';
 $string['lockwaitnotification'] = 'The majority of the duration was spent waiting for a session lock, the page may not be slow.';
+$string['long_interval_s'] = 'Processing interval (seconds)';
+$string['long_interval_s_desc'] = 'Checks the current status of long running tasks every N seconds and processes as required.
+    This includes saving profiles of finished cron tasks and partial saves of ongoing web processes if enabled.';
```
bwalkerl: These look to be in the wrong commit - they were removed as part of the CI cleanup but added back here?

classes/sample_set.php
```diff
@@ -110,32 +136,90 @@ public function add_many_samples(iterable $samples) {
...
+    }
+
+    /**
+     * Merge memory usage sample set
+     * @param array $samples
+     * @param int $distance
+     * @return array merged sample set.
+     */
+    private function merge_memory_usage_sample_set(array $samples, int $distance = 2): array {
```
bwalkerl: I don't think the term 'distance' covers the intended meaning since we're merging essentially merging groups. Maybe chunksize?

classes/sample_set.php
```diff
@@ -110,32 +136,90 @@ public function add_many_samples(iterable $samples) {
...
+            // For the case the number of sample is odd.
+            // We keep the last one as is.
+            if ($i == count($samples) - 1) {
+                $newsamples[] = $samples[$i];
+            } else {
+                // The final value should be the higher value which can highlight the considerable memory usage.
+                $max = $samples[$i]['value'];
+                for ($j = 1; $j < $distance; $j++) {
+                    $max = max($max, $samples[$i + $j]['value']);
+                }
```
bwalkerl: From this a glance this function looks functional but a bit over complicated. Rather than adding the last odd sample, can we just do something like:

```
$count = count($samples);
for ($i = 0; $i < count; $i += $chunksize) {
    $max = $samples[$i]['value'];
    $end = min($i + $chunksize, $count);
    for ($j = $i + 1; $j < $end; $j++) {
        $max = max($max, $samples[$j]['value'])
    }
```

It's still a bit messy, but we need to be careful about performance here so this may be for the best.

classes/sample_set.php
```diff
@@ -110,32 +136,90 @@ public function add_many_samples(iterable $samples) {
...
+     * @param int $distance
+     * @return array merged sample set.
+     */
+    private function merge_excimer_sample_set(array $samples, int $distance = 2): array {
+        $newsamples = [];
+        for ($i = 0; $i < count($samples); $i += $distance) {
+            // For the case the number of sample is odd.
+            // We keep the last one as is.
+            if ($i == count($samples) - 1) {
+                $newsamples[] = $samples[$i];
```
bwalkerl: See comment from memory - would be ideal to move this into an early end of the second loop.

classes/sample_set.php
```diff
@@ -110,32 +136,90 @@ public function add_many_samples(iterable $samples) {
...
+                $trace = $samples[$i]['trace'];
+                $maxcount = $samples[$i]['eventcount'];
+                $eventcount = $samples[$i]['eventcount'];
+                for ($j = 1; $j < $distance; $j++) {
+                    if ($maxcount <= $samples[$i + $j]['eventcount']) {
+                        $trace = $samples[$i + $j]['trace'];
+                        $maxcount = $samples[$i + $j]['eventcount'];
+                    }
+                    // Halving everytime (TODO: need to consider if averaging is better).
+                    $eventcount = ($eventcount + $samples[$i + 1]['eventcount']) / 2.0;
```
bwalkerl: I think eventcount needs to be the average. So likely $eventcount += $samples[j]['eventcount'] in the loop and average divide by the number of samples in the chunk afterwards (this won't always be chunksize on the last).

tests/excimer_testcase.php
```diff
@@ -172,4 +172,31 @@ protected function get_manager_stub(processor $processor, \ExcimerProfiler $prof
...
+     * @param \ExcimerLog $log
+     * @return array
+     */
+    protected function from_log_to_samples($log) {
+        $samples = [];
+        foreach ($log as $entry) {
+            $samples[] = $this->from_log_entry_to_sample($entry);
+        }
+        return $samples;
+    }
```
bwalkerl, line 201: These work, but I can't help but think there might be a simpler way to do this without so many changes.

bwalkerl (changes requested):
Thanks @duyhuynhdev 

This is a much less destructive change than the previous PR, so we should be able to get this merged in once we iron out everything. The previous PR still does have other benefits, especially for web, but is just unworkable for cron without a clearer way to know when tasks end. 

One more thing that is missing here that was in the previous PR is the changes to store event count as well (and fixes for import). I think that would be useful to have here.

duyhuynhdev:
Hi Ben, I have updated the code. Could you review it again when you back. Thanks

classes/sample_set.php
```diff
@@ -75,27 +75,50 @@ public function get_stack_depth(): int {
...
+     * @param bool $ismemory
+     */
+    private function get_sample_from_pool(bool $ismemory = false): array {
+        if ($ismemory) {
+            $mergedsample = $this->merge_memory_usage_sample_set($this->samplepool, $this->filterrate);
+        } else {
+            $mergedsample = $this->merge_excimer_sample_set($this->samplepool, $this->filterrate);
+        }
+        $this->samplepool = [];
+        return $mergedsample[0];
```
bwalkerl: The use of [0] at the end should have a comment as it isn't immediately obvious that there's only 1 value because the size of the samplepool matches the filter.

Maybe `reset($mergedsample);` would be cleaner.

lang/en/tool_excimer.php
```diff
@@ -60,195 +59,158 @@
...
 $string['field_maxlockheld'] = 'Longest';
+$string['field_memoryusagemax'] = 'Max Memory Used';
+$string['field_mincreated'] = 'Earliest';
+$string['field_minduration'] = 'Fastest';
 $string['field_minlockheld'] = 'Shortest';
-$string['field_requestcount'] = 'Num profiles';
+$string['field_month'] = 'Month';
+$string['field_name'] = 'Name';
+$string['field_numsamples'] = 'Number of samples';
+$string['field_numsamples_value'] = '{$a->samples} samples ({$a->events} events) @ ~{$a->samplerate}ms';
```
bwalkerl: Two things:
1) $a->events isn't being passed so it currently doesn't exist
2) We need a clearer way to show this in the UI, either by explaining the difference with a help field or just showing events. The graph also needs to be consistent with terminology. Maybe this should just be `{$a->events} events @  ~{$a->samplerate}ms`? If you change this make sure to default back to using samples if events aren't stored for backwards compatibility.

classes/sample_set.php
bwalkerl: Strictly speaking it should work fine after your change to swap the order, but I think the doubling makes more sense where it was before this PR at the start of the function.

classes/web_processor.php
```diff
@@ -149,8 +149,9 @@ public function process(manager $manager, bool $isfinal) {
             $this->profile->set('reason', $reason);
             $this->profile->set('finished', $isfinal ? (int) $current : 0);
             $this->profile->set('memoryusagedatad3', $this->memoryusagesampleset->samples);
-            $this->profile->set('flamedatad3', flamed3_node::from_excimer_log_entries($this->sampleset->samples));
-            $this->profile->set('numsamples', $this->sampleset->count());
+            $this->profile->set('flamedatad3', flamed3_node::from_sample_set_samples($this->sampleset->samples));
+            $this->profile->set('numsamples', count($this->sampleset->samples));
+            $this->profile->set('numevents', $this->sampleset->count());
```
bwalkerl: The duration in the graph isn't quite matching up. In my example, the duration is 1 min 6 secs but the graph is showing a duration of 1 min 2 secs. This is with 737 samples and 772 events @ ~80ms.

Need to confirm whether this is using the right values to calculate that.

classes/sample_set.php
```diff
@@ -110,32 +133,84 @@ public function add_many_samples(iterable $samples) {
...
+                // Conduct sum and count for averaging.
+                $sumeventcount += $samples[$j]['eventcount'];
+                $eventcount++;
+            }
+            $newsamples[] = [
+                'eventcount' => (int)($sumeventcount / $eventcount),
+                'trace' => $trace,
+            ];
+        }
+        return $newsamples;
```
bwalkerl: The duration in the graph isn't quite matching up. In my example, the duration is 1 min 6 secs but the graph is showing a duration of 1 min 2 secs. This is with 737 samples and 772 events @ ~80ms.

We don't expect it to be 100% accurate after the merging but 4s here is fairly significant. This function could possibly use some more detailed unit tests as well for higher chunk sizes.

bwalkerl (changes requested):
This is looking close, the main issue now is just the discrepancy with the duration in the graph, possibly due to a bug in merging or other code needing to be updated. 

I've a few new comments alongside the previous ones.

classes/sample_set.php
bwalkerl: This looks to because of the casting to (int) for eventcount. After removing that it looks like it just works, with implicit conversion warnings later down the line. I think what we need to do is leave it as a float at this point and explicitly have the conversion later in from_sample_set_samples() and count(). 

Make sure the reasoning is commented.

classes/sample_set.php
```diff
@@ -110,32 +133,84 @@ public function add_many_samples(iterable $samples) {
...
+     */
+    private function merge_excimer_sample_set(array $samples, int $chunksize = 2): array {
+        $newsamples = [];
+        $count = count($samples);
+        for ($i = 0; $i < $count; $i += $chunksize) {
+            // We keep stacktrace of the bigger sample as it contains the information of long-running task.
+            $trace = $samples[$i]['trace'];
+            $maxeventcount = $samples[$i]['eventcount'];
+            $sumeventcount = $samples[$i]['eventcount'];
+            $eventcount = 1;
```
bwalkerl: 1) It would be cleaner to handle these inside the loop and start at $j = $i. 
2) We shouldn't use $eventcount as the name for the number of chunks in this loop. $tracecount or $chunkcount would be better. This can also be pre-calculated with `$tracecount = $end - $i;`.

classes/sample_set.php
bwalkerl: After some more testing I'm not sure whether it is the casting to (int).

classes/sample_set.php
duyhuynhdev: solved

lang/en/tool_excimer.php
duyhuynhdev: I'll keep the old one (w/o events)

classes/sample_set.php
duyhuynhdev: I am testing with float values to see if the issue is still there.

classes/sample_set.php
duyhuynhdev: Thanks, I updated

classes/flamed3_node.php
```diff
@@ -69,18 +69,16 @@ public function add_excimer_trace_tail(array $tail, int $eventcount): void {
...
      */
-    public static function from_excimer_log_entries(iterable $entries): flamed3_node {
+    public static function from_sample_set_samples(array $samples): flamed3_node {
         $root = new flamed3_node('root');
-        foreach ($entries as $entry) {
-            $eventcount = $entry->getEventCount();
-            $trace = array_reverse($entry->getTrace());
-            $root->add_excimer_trace_tail($trace, $eventcount);
+        foreach ($samples as $sample) {
+            $root->add_excimer_trace_tail(array_reverse($sample['trace']), (int) $sample['eventcount']);
```
bwalkerl: This is still being cast to int here when the function accepts floats now.

classes/task_sample_set.php
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Task sample set.
+ *
+ * @package    tool_excimer
+ * @author     Dustin Huynh <dustinhuynh@catalyst-au.net>
+ * @copyright  2026 Catalyst IT
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class task_sample_set extends sample_set {
```
bwalkerl: Technically speaking, this isn't a task sample set, as it includes ones from web. `excimer_sample_set` would be a more accurate name.

lang/en/tool_excimer.php
```diff
@@ -99,8 +100,8 @@
 $string['field_minlockheld'] = 'Shortest';
 $string['field_month'] = 'Month';
 $string['field_name'] = 'Name';
-$string['field_numsamples'] = 'Number of samples';
-$string['field_numsamples_value'] = '{$a->samples} samples ({$a->events} events) @ ~{$a->samplerate}ms';
+$string['field_numevents'] = 'Number of events';
+$string['field_numevents_value'] = '{$a->events} events @ ~{$a->samplerate}ms';
```
bwalkerl: We're still measuring samples here, so field_numsamples is still the right name. If we're only including one value, I would keep it as:
`$string['field_numsamples_value'] = '{$a->events} events @ ~{$a->samplerate}ms';`

And leave the tag as 'Number of samples.'

templates/flamegraph.mustache
```diff
@@ -59,7 +59,8 @@ init();
 window.addEventListener('resize', _ => init());
 
 function displaysamplestats(d) {
-    let time = d.value * {{samplerate}};
+    let count = parseInt(d.value);
+    let time = count * {{samplerate}};
```
bwalkerl: Time should use the float to be more accuracy

classes/sample_set.php
```diff
@@ -75,27 +75,37 @@ public function get_stack_depth(): int {
...
+     */
+    abstract public function process_samples($sample);
+
+    /**
+     * Merging sample set
+     *
+     * @param array|\ExcimerLogEntry $samples
+     * @param int $chunksize
+     */
+    abstract public function merge_sample_set($samples, int $chunksize = 2);
```
bwalkerl: Both functions treat samples as arrays - can we do the same here?

classes/cron_processor.php
```diff
@@ -126,8 +126,8 @@ public function on_interval(manager $manager, bool $isfinal = false) {
 
             // If there exists a current task, and the sampleset for it is not created yet, create it.
             if ($taskname && ($this->tasksampleset === null)) {
-                $this->tasksampleset = new sample_set($taskname, $this->sampletime);
-                $this->memoryusagesampleset = new sample_set($taskname, $this->sampletime);
+                $this->tasksampleset = new task_sample_set($taskname, $this->sampletime);
+                $this->memoryusagesampleset = new memory_sample_set($taskname, $this->sampletime);
```
bwalkerl: We should also update the var type for $this->tasksampleset to the specific class.

bwalkerl (commented):
Thanks @duyhuynhdev 

The sample set abstract class looks great and we've solved the biggest issue, just a few more things to tidy up.

classes/sample_set.php
```diff
@@ -24,28 +24,28 @@
...
+    /** @var array An array of sample objects. It could contain memory usage samples or tasks samples  */
     public $samples = [];
 
+    /** @var array An array of sample objects. It is used to store sample upto the filterrate  */
+    public $samplepool = [];
+
     /** @var int Sample limit. */
     public $samplelimit;
     /** @var int The maximum stack depth. */
     public $maxstackdepth = 0;
```
bwalkerl: Max stack depth is specific to excimer sample sets, so we should be able to move that and get_stack_depth(): to the new class.

tests/excimer_testcase.php
duyhuynhdev: I also have no idea how to make it simpler ^^. Let me think about after solving all main issues

bwalkerl (approved):
This looks to be working as intended from my testing. 

There can still be edge cases such as two samples with a large number of events next to each other, but this should be much better than blinding dropping half. 

Happy for this to be merged, as long as we are OK with the terminology:
* One thing that can be discussed further is the terminology differences between sample counts and event counts. We have replaced sample counts with event counts, and stored the actual sample counts separately. This does look messy. Would we be better off keeping eventcounts as numsamples and add a new numlogentires column instead?

bwalkerl (changes requested):
@duyhuynhdev After thinking about it some more I think we have the update the terminology for consistency. 

- Leave numsamples alone, and change the UI back to 'samples' instead of 'events'. This will be the eventcount, as it was before this update. We don't add numevents.
- Instead of storing the number of log entries as numsamples, store it as numlogentires or numentries
- There was a fallback method when numevents doesn't exist that can be removed.

This would keep consistency with the terminology in the UI and tables, and should reduce confusion.

db/upgrade.php
```diff
@@ -560,5 +560,15 @@ function xmldb_tool_excimer_upgrade($oldversion) {
         upgrade_plugin_savepoint(true, 2025120903, 'tool', 'excimer');
     }
 
+    if ($oldversion < 2026022500) {
+        $table = new xmldb_table('tool_excimer_profiles');
+        // Add 'numlogentries' field - The number of events taken.
+        $field = new xmldb_field('numlogentries', XMLDB_TYPE_INTEGER, '11', true, XMLDB_NOTNULL, null, 0, 'numsamples');
```
bwalkerl: It looks like this wasn't generated with XMLDB editor - it has schema alignment issues with the default (likely should be null) and possibly some other arguments.

bwalkerl (approved):
Thanks @duyhuynhdev this is looking good to me now, matches the spec, and the confusion has been removed so happy for it to be merged.

The only thing left I can see that might be a concern is the size of the samplepool when we reach high filterrates, but that can be mitigated by settings so I'm not sure how much of a concern it is.

PR #428, opened 2026-04-16
Fix sql dml_read_exception with space after COUNT function

kristian-94:
The grouped profile table pages (slowest_web.php, slowest_cli.php, etc.) throw a dml_read_exception on MySQL
without IGNORE_SPACE in sql_mode, which is the default (STRICT_ALL_TABLES is set by core):

```
FUNCTION dbname.COUNT does not exist.
Check the 'Function Name Parsing and Resolution' section in the Reference Manual
```

### Test

The test now runs the table->out() function, and without this fix has an error in default mysql 9.6 and 8.X: 

```
vendor/bin/phpunit --filter tool_excimer_grouped_profile_table_test
Moodle 4.5.8+ (Build: 20260116), 3c33567cad20e60a8382ea3e9e3b1f59a4bde5f5
Php: 8.3.30, mysqli: 9.6.0, OS: Darwin 25.4.0 arm64
PHPUnit 9.6.29 by Sebastian Bergmann and contributors.

E.                                                                  2 / 2 (100%)

Time: 00:00.826, Memory: 410.00 MB

There was 1 error:

1) tool_excimer\tool_excimer_grouped_profile_table_test::test_grouped_script_profile_table_query_executes
dml_read_exception: Error reading from database (FUNCTION test.COUNT does not exist. Check the 'Function Name Parsing and Resolution' section in the Reference Manual
```

One line fix to remove that space, thanks!

### DIFF
--- a/tests/tool_excimer_grouped_profile_table_test.php
+++ b/tests/tool_excimer_grouped_profile_table_test.php
@@ -1,0 +1,101 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_excimer;
+
+defined('MOODLE_INTERNAL') || die();
+
+require_once(__DIR__ . "/quick_save_trait.php");
+
+/**
+ * Tests for grouped_profile_table SQL generation.
+ *
+ * @package    tool_excimer
+ * @copyright  2026 Catalyst IT
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ * @covers     \tool_excimer\grouped_profile_table
+ */
+final class tool_excimer_grouped_profile_table_test extends excimer_testcase {
+    use quick_save_trait;
+
+    /**
+     * Set up before each test.
+     */
+    protected function setUp(): void {
+        parent::setUp();
+        $this->resetAfterTest();
+        profile_helper::init();
+        script_metadata::init();
+    }
+
+    /**
+     * Test that the grouped script profile table query executes without SQL errors.
+     *
+     * Regression test: COUNT (CASE ...) with a space between COUNT and the
+     * opening parenthesis caused "FUNCTION dbname.COUNT does not exist" on
+     * MySQL servers without the IGNORE_SPACE sql_mode.
+     */
+    public function test_grouped_script_profile_table_query_executes(): void {
+        $node = new flamed3_node('root', 0);
+
+        $this->quick_save('/course/view.php?id=1', $node, profile::REASON_SLOW, 2.5);
+        $this->quick_save('/course/view.php?id=2', $node, profile::REASON_SLOW, 1.0);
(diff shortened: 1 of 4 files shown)


tests/tool_excimer_grouped_profile_table_test.php
```diff
@@ -0,0 +1,130 @@
...
+ * @covers     \tool_excimer\grouped_profile_table
+ */
+final class tool_excimer_grouped_profile_table_test extends excimer_testcase {
+
+    /**
+     * Set up before each test.
+     */
+    protected function setUp(): void {
+        parent::setUp();
+        $this->preventResetByRollback();
```
marxjohnson: What is the reason for `preventResetByRollback()`? Check that this is definitely necessary, if it is please add a comment explaining why.

tests/tool_excimer_grouped_profile_table_test.php
```diff
@@ -0,0 +1,130 @@
...
+     *
+     * @param string $request
+     * @param flamed3_node $node
+     * @param int $reason
+     * @param float $duration
+     * @param int $created
+     * @param string $lockreason
+     * @return int The ID of the record.
+     */
+    public function quick_save(
```
marxjohnson: This function looks like it's been copied from `tool_excimer_profile_helper_test.php`. Rather than duplicating it, it would be better to move it to a trait which can be used by both test classes.

tests/tool_excimer_grouped_profile_table_test.php
```diff
@@ -0,0 +1,130 @@
...
+        $table = new grouped_script_profile_table('test_grouped_script');
+        $table->set_scripttypes([profile::SCRIPTTYPE_WEB, profile::SCRIPTTYPE_AJAX]);
+
+        $url = new \moodle_url('/admin/tool/excimer/slowest_web.php');
+        $table->set_url_path($url);
+        $table->sortable(true, 'maxduration', SORT_DESC);
+        $table->define_baseurl($url);
+        $table->make_columns();
+
+        ob_start();
```
marxjohnson: Instead of capturing the output like this, you can use `$this->expectOutputRegex()` to properly assert that output is produced and contains what you expect.

marxjohnson (changes requested):
@kristian-94 Thanks for the contribution. The fix itself looks OK, but there are a few points in the unit tests that I think could be improved, please see my comments.

marxjohnson:
@kristian-94 Also, please see the errors from codechecker.

tests/tool_excimer_grouped_profile_table_test.php
kristian-94: No real reason, just following the pattern of what the existing tests already do. Have removed it

tests/tool_excimer_grouped_profile_table_test.php
kristian-94: Sure, done

kristian-94: Thanks, done

tests/tool_excimer_grouped_profile_table_test.php
kristian-94: sure, done

kristian-94:
Hi @marxjohnson , let me know what you think now,  I've run the checks locally and they should also be passing, thanks

kristian-94:
Hi @marxjohnson , did you see this?

marxjohnson:
@kristian-94 Thanks for the nudge. It looks like the same fix was already applied by @brendanheywood, but I think it's worth having the extra test coverage you've provided, so once all the checks come back green I will merge this.

brendanheywood:
ah sorry @kristian-94 I didn't even think to check +1 to merge with more tests (I did think I should add some after I did that but I was moving pretty fast...)

kristian-94:
@marxjohnson I can't run the tests. Even if I rebase this you will have to click run manually right? So I will have to ping you again and wait for that

marxjohnson:
@brendanheywood It looks like the tests have run in the Actions tab, but it still says its waiting in 8 checks here, do you know what its waiting for?

brendanheywood:
Nope no idea, merged anyway

PR #433, opened 2026-07-08
WR490986: Recommended amendments.

vladaskidanovas:


classes/check/slowest.php
```diff
@@ -1,5 +1,5 @@
 <?php
-// This file is part of Moodle - http://moodle.org/  <--change
+// This file is part of Moodle - https://moodle.org/
```
cwarwicker: You don't need to make changes to the code moodle header comment. It's the same on all files. So the changes here and in any other files where the core header comment is changed, you don't need those.

classes/local/hooks/after_config.php
```diff
@@ -40,7 +40,7 @@ class after_config {
      * See also https://docs.moodle.org/dev/Login_callbacks#after_config.
      *
      * @param \core\hook\after_config $hook
-     * return void
+     * @return void
```
cwarwicker: Having a return void in the docblocks is actually discouraged. So, where you've changed it here, or added it in other bits, it's not needed. As long as the method itself returns void, it's considered unncecessary to put a void return in the docblock as well.

cwarwicker:
Can you also check the failing jobs. E.g. the version bump one needs you to increment the version.

cwarwicker:
Approved

PR #434, opened 2026-08-12
Bug fixes and improvements

michaelkotlyar:


cwarwicker:
Looks ok. As long as tests pass and you've tested locally and nothing breaks.

classes/grouped_script_profile_table.php
```diff
@@ -40,7 +40,7 @@ protected function get_group_by(): string {
      * @return string
      */
     public function col_scriptgroup(\stdClass $record): string {
-        $displayedvalue = $record->scriptgroup;
+        $displayedvalue = format_text($record->scriptgroup, FORMAT_PLAIN);
```
brendanheywood, line 43: why are you wanting to format this? Are you actually wanting to make sure it is escaped correctly?

classes/page_group_table.php
```diff
@@ -100,8 +100,9 @@ public function setup() {
      * @return string
      */
     public function col_name(\stdClass $record): string {
+        $displayedvalue = format_text($record->name, FORMAT_PLAIN);
```
brendanheywood, line 103: ditto, this feels like an escaping this not a formatting thing

classes/profile.php
```diff
@@ -259,10 +259,10 @@ public function add_env(string $request): void {
         // Store the sample rate at the time this profile is created.
         $this->raw_set('samplerate', script_metadata::$samplems);
 
-        $this->raw_set('method', $_SERVER['REQUEST_METHOD'] ?? '');
-        $this->raw_set('pathinfo', $_SERVER['PATH_INFO'] ?? '');
-        $this->raw_set('useragent', $_SERVER['HTTP_USER_AGENT'] ?? '');
-        $this->raw_set('referer', $_SERVER['HTTP_REFERER'] ?? '');
+        $this->raw_set('method', clean_param($_SERVER['REQUEST_METHOD'] ?? '', PARAM_ALPHA));
```
brendanheywood, line 262: why do you want to clean this in the raw data?

I can understand escaping it when displaying it if that is not already happening, but cleaning the data reduces the usefulness of the raw metadata for debugging

michaelkotlyar: Storing this raw data is supposedly a non-risk as long as we output it correctly and you make a fair point about this limitting debugging. But since anyone can create an entry by adding FLAMEME=1 at the end of a request, I think there is a reasonable concern here. Perhaps instead of cleaning on entry, we limit flaming to admins?

brendanheywood: Anyone can create an entry inadvertently without flameme for other reasons, so that should not be a factor. We should store the raw value the same as an access log file would, we just need to ensure it's escaped on output.

